PR #11, opened 2017-04-05
update gradle

wclausen:
Android Studio 2.3 keeps prompting me to do this and evidently instant run won't work unless this change is made.

Is there some reason we can't just bump?


fabiendevos:
Nope, sounds good!

PR #29, opened 2017-04-26
Added inflate method in BaseScreenView that does not take context as …

theyann:
…parameter

I have mental issues and my OCD sometimes takes the best of me. In this case, this wrapper inflate method will simply use the getContext(), so that the view code will be even simpler and clean :)

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/BaseScreenView.java
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/BaseScreenView.java
@@ -27,6 +27,6 @@ public class BaseScreenView<S extends Screen> extends FrameLayout implements Scr
   }
 
   public View inflate(int resource, ViewGroup root) {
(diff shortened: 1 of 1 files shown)


theyann:
This is the last time I do a PR directly from GitHub ... I don't handle imports very well without my trusty IDE ;)

magellan-library/src/main/java/com/wealthfront/magellan/BaseScreenView.java
```diff
@@ -24,4 +26,7 @@ public final S getScreen() {
     return screen;
   }
 
+  public View inflate(int resource, ViewGroup root) {
+    return inflate(getContext(), resource, root);
```
fabiendevos: Shouldn't the root be "this"?
Should be the common case, and makes it truly simpler in impl

fabiendevos:
Thanks @theyann !!!

PR #33, opened 2017-05-11
Screen setTitle takes a CharSequence

kyeungk:
Make it so setTitle accepts a string on top of the existing @StringRes


magellan-library/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -204,6 +204,10 @@ protected final void setTitle(@StringRes int titleResId) {
     activity.setTitle(titleResId);
   }
 
+  protected final void setTitle(String title) {
```
FabianTerhorst: Should be ```CharSequence title```

FabianTerhorst:
This is an required feature. Otherwise it would not be possible to use string formats. For example:
```setTitle(getString(R.string.title, name));```

PR #34, opened 2017-05-14
Add Toolbar Example to Advanced Sample App

wclausen:
Toolbar example includes:

- changing action bar color when navigating
- hiding/showing action bar when navigating
- use of options menu/overflow menu

toolbar_example


dimparf:
Wow, thanks!

fabiendevos:
This is pretty good as is, but do we want to polish it a bit more before merging?
Status bar color for example. Also maybe we want to "show()" when going to the actionbar-less screen to make it a nicer transition?

PR #35, opened 2017-05-18
@ParametersAreNonnullByDefault

jmfayard:
Hello, Magellan is awesome, but when I use it from Kotlin I have to correct each function to state that the parameters it gives are actually non-nullable.

It would be great if we can implement the solution from
https://medium.com/square-corner-blog/non-null-is-the-default-58ffc0bb9111

By adding `@ParametersAreNonnullByDefault`, `convert java file to kotlin` now knows that context is actually non-nullable

```kotlin
internal class DetailScreen : Screen<DetailView>() {
    override fun createView(context: Context): DetailView {
        return DetailView(context)
    }
    override fun getTitle(context: Context): String {
        return "Detail Screen"
    }
}
```

My PR is intended as a starting point, I did not check if some methods in the library needs `@Nullable`, and you maybe don't want to add a kotlin dependancy to the sample.


fabiendevos:
@jmfayard thanks a lot for the starting point! We are definitely going to do this! Yes we need to go through all the methods and apply @Nullable when necessary.

jmfayard:
Hello @fabiendevos 

so I was too early in May because good support for jsr-305 is only available since kotlin 1.1.4, but since it's now available it would totally make sense to introduce `@ParametersAreNonnullByDefault` and something like Spring's `@NonNullApi`

some of the most important type hints I would like to see in practice:

image


References:

https://blog.jetbrains.com/kotlin/2017/08/kotlin-1-1-4-is-out/

https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/lang/NonNullApi.java

https://github.com/Kotlin/KEEP/blob/41091f1cc7045142181d8c89645059f4a15cc91a/proposals/jsr-305-custom-nullability-qualifiers.md

jmfayard:
Status update: I implemented the code, but it's currently useless because of an Android Studio bug that I reported here [#68304819](https://issuetracker.google.com/issues/68304819)
So waiting for it to be fixed

PR #46, opened 2017-06-22
Leak canary integration example in magellan. It needs screen to have …

bangarharshit:
…non final onHide method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /build
 /captures
 .externalNativeBuild
+.idea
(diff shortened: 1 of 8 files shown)


fabiendevos:
I think your realized it but yeah you should use onUnsubscribe instead of onHide in an RxScreen.

bangarharshit:
@fabiendevos onUnSubscribe is also incorrect. It always shows screen leaked since it thinks that screen is collected after onUnSubscribe which is not true. There is no callback such as onDestroy for the screen pop.

Another question  - How do you design network call. If I made a network call in onShow of a screen and go to another screen and come back it is done again.

fabiendevos:
Good point. We need a proper integration with leakcanary for sure. 

For network call either you need to be ok redoing the call (for example because it's a small one or because you have a cache) otherwise extract it in a different object like a Singleton.

PR #51, opened 2017-07-05
Updated inflate statement for the basic sample in README file

dharmin007:
Updated the README file to reflect the sample code from the basic sample app.


fabiendevos:
Thanks!. Actually, it's the other way around though: the sample should be changed to make use of this convenience function :)

dharmin007:
Thanks for checking Fabien. I'll close this pull request.

PR #58, opened 2017-10-02
Update README.md

StefMa:
This is just a very minor but it looks way better IMO.
Also I think that most of the developers out there are more familiar with that "style" for dependencies...

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ compile 'com.wealthfront:magellan:1.0.0'
 
 ```gradle
 def magellanVersion = '1.0.0'
-compile 'com.wealthfront:magellan:' + magellanVersion
-compile 'com.wealthfront:magellan-support:' + magellanVersion
-compile 'com.wealthfront:magellan-rx:' + magellanVersion
+compile "com.wealthfront:magellan:$magellanVersion"
+compile "com.wealthfront:magellan-support:$magellanVersion"
+compile "com.wealthfront:magellan-rx:$magellanVersion"
 ```
 
 ### Add-on coming soon


fabiendevos:
I actually find this slightly harder to read... 
I couldn't find examples on other libs. Do you have any example in mind?

StefMa:
Unfortunately. No.
I can't find an example expect of [ThirtyInch](https://github.com/grandcentrix/thirtyinch#get-it--) 🙃 

In our company we do everything with the syntax I apply here.
Eg.
```gradle
    //Retrofit
    implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
    implementation "com.squareup.retrofit2:converter-moshi:$retrofitVersion"
    implementation "com.squareup.retrofit2:adapter-rxjava2:$retrofitVersion"

    //Support Lib
    implementation "com.android.support:appcompat-v7:$supportLibVersion"
    implementation "com.android.support:support-v4:$supportLibVersion"
    implementation "com.android.support:design:$supportLibVersion"
    implementation "com.android.support:cardview-v7:$supportLibVersion"
```

I find it way better to read. 
But not only Groovy brings String Template. Even [Kotlin does it](https://kotlinlang.org/docs/reference/basic-types.html#string-templates). 
For me the String concat with a plus (`+`) cames from the old Java days... And even there are [some helpers](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html) which try to bring String templates to it. 

But if you don't like it. Feel free to close that PR 🙃

fabiendevos:
@kimingy @wclausen @narenkmanoharan @ryanmoelter what do you think? I slightly prefer the "+" version

PR #70, opened 2018-02-26
Save Transitions to use for back navigation

ryanmoelter:
Saving transitions allow us to create interesting transitions later, in addition to just making sense. I'd rather not have to override back transitions too.

Saving transitions can be tricky when you can rewrite the backstack, so I decided to attach Transitions to screens so they can animate out in the same method they animate in. A mutable backstack in this case means that transitions must work with any to/from view, which is a reasonable restriction in my mind.

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
@@ -548,8 +548,18 @@ public class Navigator implements BackHandler {
   private void animateAndRemove(
       final View from, final View to, final NavigationType navType, final Direction direction) {
     ghostView = from;
-    final Transition transitionToUse = overridingTransition != null ? overridingTransition : transition;
-    overridingTransition = null;
+    Transition currentTransition = transition;
+    if (overridingTransition != null) {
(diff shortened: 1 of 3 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
```diff
@@ -548,8 +548,18 @@ private void navigate(final Direction direction, final NavigationType navType, f
...
       final View from, final View to, final NavigationType navType, final Direction direction) {
     ghostView = from;
-    final Transition transitionToUse = overridingTransition != null ? overridingTransition : transition;
-    overridingTransition = null;
+    Transition currentTransition = transition;
+    if (overridingTransition != null) {
+      currentTransition = overridingTransition;
+      overridingTransition = null;
+    } else if (direction.equals(Direction.BACKWARD) && currentScreen().getLeaveTransition() != null) {
+      currentTransition = currentScreen().getLeaveTransition();
```
ryanmoelter: I want to test this logic, but triggering the `whenMeasured` block is tough. I've tried manually calling `measure` and `layout`, but it's still not triggering it. Thoughts/suggestions?

PR #71, opened 2018-03-07
Support activity title modification

cmathew:
notify navigator when the data backing its title has changed. 

For example, if the formula for your activity title is something like `user.getName()`, you'd likely want the title to be refreshed each time the user's name changes.


cmathew:
I assumed that developers would only be interested in updating the *current* Screen's title, hence this approach is Navigator-centric. Would a Screen-centric one be better?

PR #73, opened 2018-03-16
Add nullability in magellan

narenkmanoharan:
Add better interoperability with Kotlin by adding JSR-305 annotation.

- @ParametersAreNonnullByDefault
- @Nullable
- @NonNull

References: 

- https://medium.com/square-corner-blog/rolling-out-nullable-42dd823fbd89
- https://medium.com/square-corner-blog/non-null-is-the-default-58ffc0bb9111

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/Screen.java
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/Screen.java
@@ -12,6 +12,9 @@ import android.util.SparseArray;
 import android.view.Menu;
 import android.view.ViewGroup;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
 import static com.wealthfront.magellan.Preconditions.checkState;
 
 /**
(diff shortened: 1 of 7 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -143,7 +146,7 @@ protected boolean shouldAnimateActionBar() {
     return true;
   }
 
-  public String getTitle(Context context) {
+  public String getTitle(@Nonnull Context context) {
```
fabiendevos: why do we need the @Nonnull s since we have @ParametersAreNotNullByDefault?

ryanmoelter:
Do we want to make these nullability annotations a warning to allow people to migrate slowly?

fabiendevos:
@narenkmanoharan can you check what's breaking?

jmfayard:
@fabiendevos I've lot of experience with Magellan and I would say makes this a breaking change and add a small guide on how to upgrade.
The breaking changes are things we really needs to take into account to have a stable app (always check that the activity and the view is not null)

PR #75, opened 2018-04-02
Added ScreenAwareActivity to add flexibility to use other Activities.

jmholtan:
Implemented an interface called ScreenAwareActivity that adds flexibility to library consumers. This change allows a developer to not be forced into using Activity, it changes the system to rely on Context and not Activity. This is a breaking change as it requires developers to implement the new interface.

I made this change because I was unable to use an Activity in a project but still wanted to use the Magellan library.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,8 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="EntryPointsManager">
-    <entry_points version="2.0" />
-  </component>
   <component name="NullableNotNullManager">
     <option name="myDefaultNullable" value="android.support.annotation.Nullable" />
     <option name="myDefaultNotNull" value="android.support.annotation.NonNull" />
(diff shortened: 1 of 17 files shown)


fabiendevos:
Hi @jmholtan ,

What do you mean by "unable to use an Activity"? You definitely need an activity to display UI on Android unless all you implement is a Service, in which case you don't need Magellan.

Can you explain more about the exact situation?

We cannot take breaking changes without bumping the version to 2.0, which we might not want to do quite yet.

jmholtan:
Hi @fabiendevos,

Apologies on the vagueness of the the 'unable to use an Activity'. I am working on an Android Auto application. I am unable to get into many specifics but there is a different Activity in my case which does not extend Activity. Hence the need to switch to an interface style structure. 

Hopefully that gives you a bit more insight into the reasoning. I understand if these changes can not be accepted at this time. I still wanted to propose the solution to you and your team.

Thanks!

fabiendevos:
> A different Activity which does not extend Activity

Uh? I would need to know your architecture to help. What class does this extend exactly?

jmholtan:
Apologies but I have to close this as I am unable to discuss further.

PR #78, opened 2018-05-08
Change default transition

ryanmoelter:
The default transition handles almost all of the normal transition styles, but it uses the default `AccelerateDecelerateInterpolator`, whereas material design guidelines recommend the `FastOutSlowInInterpolator`. Unfortunately, the `FastOutSlowInInterpolator` is part of the support library, which would be too large to include in Magellan.

I propose we add the ability to set the interpolators for the default transition to support this and other cases. Also, I propose that we allow different interpolators for `NavigationType.GO` vs. `SHOW` vs. `HIDE`, since they are all different animations.

(Note: it looks like they're splitting the interpolators out into a separate package in support library v28, but that's in alpha right now.)

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/transitions/DefaultTransition.java
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/transitions/DefaultTransition.java
@@ -8,6 +8,8 @@ import android.animation.TimeInterpolator;
 import android.util.Property;
 import android.view.View;
 import android.view.animation.AccelerateDecelerateInterpolator;
+import android.view.animation.AccelerateInterpolator;
+import android.view.animation.DecelerateInterpolator;
 
 import com.wealthfront.magellan.Direction;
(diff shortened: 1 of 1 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/transitions/DefaultTransition.java
```diff
@@ -15,6 +19,25 @@
...
+
+  public DefaultTransition() { }
+
+  public DefaultTransition(TimeInterpolator customInterpolator) {
+    this.customGoInterpolator = customInterpolator;
+    this.customShowInterpolator = customInterpolator;
+    this.customHideInterpolator = customInterpolator;
+  }
+
+  public DefaultTransition(TimeInterpolator customGoInterpolator, TimeInterpolator customShowInterpolator,
```
fabiendevos: Please use nested constructor calls instead of code duplication

(previous constructor should call this one using this()

PR #102, opened 2019-02-28
Show/hide observability

cmathew:
depending on internal checks, these methods will either navigate or no-op. Add a return value to indicate which of these happened


magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
```diff
@@ -349,8 +349,8 @@ public void replaceNow(Screen screen) {
    *
    * @param screen  new top screen on back stack
```
fabiendevos: please update the javadoc everywhere to indicate what the boolean means.

PR #104, opened 2019-07-10
Add screen name to preconditions

narenkmanoharan:
Add screen name to preconditions when an exception is thrown for when the screen is not present in backstack

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
@@ -459,7 +459,7 @@ public class Navigator implements BackHandler {
     navigate(new HistoryRewriter() {
       @Override
       public void rewriteHistory(Deque<Screen> history) {
-        checkArgument(history.contains(screen), String.format("Can't go back to a screen (%s) that isn't in history.", screen.getClass().getName()));
+        checkArgument(history.contains(screen), String.format("Can't go back to a screen (%s) that isn't in history. \nHistory: (%s)", screen.getClass().getName(), history));
         while (history.size() > 1) {
           if (history.peek() == screen) {
             break;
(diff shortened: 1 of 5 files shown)


magellan-library/src/test/java/com/wealthfront/magellan/ScreenTest.java
```diff
@@ -149,20 +148,4 @@ public void whenTransitionFinished_afterTransitionFinished() {
     screen.whenTransitionFinished(listener);
     verify(listener).onTransitionFinished();
   }
-
-  private static class DummyScreen extends Screen<BaseScreenView<DummyScreen>> {
```
cmathew: nice pulling this out for re-use. Minor: looks like the same code exists in ScreenGroupTest.java if you're willing to pull it out of there as well?

magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
```diff
@@ -459,7 +459,7 @@ public void goBackTo(final Screen screen, NavigationType navigationType) {
     navigate(new HistoryRewriter() {
       @Override
       public void rewriteHistory(Deque<Screen> history) {
-        checkArgument(history.contains(screen), "Can't go back to a screen that isn't in history.");
+        checkArgument(history.contains(screen), String.format("Can't go back to a screen (%s) that isn't in history.", screen.getClass().getName()));
```
cmathew: do we want to include the current history as well?

magellan-library/src/test/java/com/wealthfront/magellan/ScreenTest.java
narenkmanoharan: d

magellan-library/src/main/java/com/wealthfront/magellan/Navigator.java
narenkmanoharan: d

PR #108, opened 2020-06-29
Bump versions and migrate to AndroidX

narenkmanoharan:
Bumping all relevant libraries and migrate to AndroidX. This PR also adds source and target compatibility for 1.8 so this might make it so that a few consumers of the library might not be able to user. But since API 30 there is a new warning about compatibility with java 1.7 and lower so just bumping it up. Let me know anyone feels strongly against it.

Also updated the travis CI job as well with the updated dependencies.


.travis.yml
```diff
@@ -4,12 +4,12 @@ android:
   components:
     - tools
     - platform-tools
-    - build-tools-27.0.3
-    - android-27
+    - build-tools-29.0.2
+    - android-30
```
ryanmoelter, line 8: `android-30` with `build-tools-29.*`? Are there `build-tools-30.*` or no?

ryanmoelter: Also, should we use `BUILD_TOOLS_VERSION` here?

ryanmoelter: Lol ignore me, this is in the travis file

narenkmanoharan: build-tools-30 is in RC so i did not wanna use that yet. But can upgrade to that later. :)

PR #109, opened 2020-06-30
Add latest kotlin plugin

narenkmanoharan:


narenkmanoharan:
> Does this mean we won't be using the `Libs`/`Version` pattern we have in other repos, or is this just the first step in that direction?

I haven't decided yet. But reading more about `buildSrc`, looks like it is very unoptimized since the configuration phase of gradle is not cached and is run every single time a task is triggered. So keeping this for now, but might integrate that in the future if managing dependencies is a pain. So far, its been fine.

PR #112, opened 2020-07-09
Magellan 2.0 prototype 

narenkmanoharan:
TODO:

- [x] OnSave/OnRestore
- [ ] Menu handling
- [ ] Dialogs
- [ ] HistoryRewriter
- [x] ScreenContainer
- [x] Improve findViewById
- [ ] Dependency chain for lifecycle dependency ordering
- [ ] Logging navigation events/Add support for global screen lifecycle listening (EventBus?)
- [ ] Add support for easier screen groups
- [ ] Add RxDisposer and CouroutineScope
- [ ] Other transitions and overriding transitions
- [ ] Adding preconditions (navigator)
- [ ] Adding tests for all classes

Sorry about the huge PR but this is the starting of a working magellan compose/flow prototype. Will be building more upon this PR and eventually breaking it down or not. We shall see.


magellan-compose/src/main/java/com/wealthfront/magellan/compose/core/Navigator.kt
```diff
@@ -26,7 +26,7 @@ import java.util.LinkedList
 
 class Navigator(private val container: (Context) -> ScreenContainer?) : LifecycleAwareComponent() {
 
-  private val backStack: Deque<Screen> = LinkedList()
+  private val backStack: Deque<Screen<*>> = LinkedList()
```
ryanmoelter, line 29: I don't think our navigator should be using `Screen`s directly, instead using navigables

PR #113, opened 2020-07-10
Move current magellan-library to magellan-legacy to make way for new magellan

narenkmanoharan:
The current plan is to move our current magellan library to be part of the legacy package. And the new composable stuff takes over the magellan-library module and jar. So in the future, consumers of the library,

New users of the library: 

```
implementation 'com.wealthfront:magellan:2.0.0'
```

Old/current users of the library:

```
implementation 'com.wealthfront:magellan:2.0.0'
implementation 'com.wealthfront:magellan-legacy:2.0.0' (Adds interop between the 1.0 & 2.0 versions)
```

We might have to pull the common classes used in both the libraries into a separate common module, but thats gonna come in a separate PR.

We will also need a migration guide for 2.0. I'll track it in this ticket


magellan-legacy/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+  compileSdkVersion 30
+  buildToolsVersion BUILD_TOOLS_VERSION
+
+  resourcePrefix 'magellan_'
+
+  defaultConfig {
+    minSdkVersion 15
+    targetSdkVersion 30
+    versionCode 1
+    versionName "1.0"
```
ryanmoelter, line 16: nit: maybe we should be using variables here?

narenkmanoharan: Yeah. We should. Since we seem to have a lot of modules, im gonna migrate to the buildSrc pattern here too. But in a separate PR.

PR #114, opened 2020-07-10
Add core lifecycle components

narenkmanoharan:
- Added LifecycleAware
- Added LifecycleOwner
- Added LifecycleRegistry
- Added LifecycleState.kt
- Added LifecycleStateMachine

### DIFF
--- a/magellan-library/src/test/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachineTest.kt
+++ b/magellan-library/src/test/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachineTest.kt
@@ -1,0 +1,172 @@
+package com.wealthfront.magellan.lifecycle
+
+import android.content.Context
+import com.wealthfront.magellan.lifecycle.LifecycleState.Created
+import com.wealthfront.magellan.lifecycle.LifecycleState.Destroyed
+import com.wealthfront.magellan.lifecycle.LifecycleState.Resumed
+import com.wealthfront.magellan.lifecycle.LifecycleState.Shown
+import org.junit.Before
+import org.junit.Test
+import org.mockito.InOrder
+import org.mockito.Mock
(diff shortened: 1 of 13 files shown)


magellan-library/src/test/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachineTest.kt
```diff
@@ -0,0 +1,172 @@
...
+import org.mockito.Mockito.verifyNoMoreInteractions
+
+class LifecycleStateMachineTest {
+
+  private val lifecycleStateMachine = LifecycleStateMachine()
+  
+  @Mock lateinit var lifecycleAware: LifecycleAware
+  @Mock lateinit var context: Context
+
+  private lateinit var inOrder: InOrder
```
ryanmoelter, line 23: This is cool!

magellan-library/src/test/java/com/wealthfront/magellan/view/LifecycleViewTest.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    assertThat(lifecycleView.view).isEqualTo(frameLayout)
+  }
+
+  @Test
+  fun onDestroyView() {
+    lifecycleView.hide(context)
+
+    assertThat(lifecycleView.view).isEqualTo(null)
+  }
```
ryanmoelter, line 37: We might want to make this test go through the whole lifecycle, declaring that the view either exists or doesn't along the way. This test doesn't test that `hide()` actually nulls the view out, for example.

narenkmanoharan: Added the `wholeLifecycle` testcase to check the state of the view in all the lifecycle callbacks.

PR #115, opened 2020-07-10
Add android jetpack lifecycle component to listen to lifecycle events

narenkmanoharan:
Integrate lifecycle jetpack component to listen to activity components


magellan-library/src/main/java/com/wealthfront/magellan/core/Navigable.kt
```diff
@@ -0,0 +1,6 @@
+package com.wealthfront.magellan.core
+
+import com.wealthfront.magellan.lifecycle.LifecycleAware
+import com.wealthfront.magellan.view.Displayable
+
+interface Navigable : LifecycleAware, Displayable
```
narenkmanoharan, line 6: Introducing `Navigable` which our screens will implement, this allows us to have multiple implementations of screens. For right now, we can have a screen which depends on view binding and another in the future for jetpack compose.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+  private val context: Activity,
+  private val containerRes: Int
+) : DefaultLifecycleObserver {
+
+  override fun onCreate(owner: LifecycleOwner) {
+    navigable.create(context)
+  }
+
+  override fun onStart(owner: LifecycleOwner) {
+    navigable.show(context)
```
narenkmanoharan, line 23: So the lifecycle is gonna work a little different in 2.0, we are gonna model exactly the same lifecycle events as the activity and react to them instead of creating our own onShow/onHide. Instead we act as if, onStart -> onShow and onStop -> on Hide, this gives us more flexibility and unifies the events in config changes and foreground/background happens.

magellan-library/src/main/java/com/wealthfront/magellan/view/Displayable.kt
```diff
@@ -0,0 +1,12 @@
+package com.wealthfront.magellan.view
+
+import android.view.View
+
+interface Displayable {
```
narenkmanoharan, line 5: All our navigables will have to implement Displayable

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
ryanmoelter: We briefly talked about introducing `onShow` as a separate event from `onStart`, to allow for a lifecycle similar to magellan's current `onShow`->`onHide`. What tipped the final decision towards the solution in this PR? Or do we want to think more about it?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.wealthfront.magellan.ScreenContainer
+
+internal class ActivityLifecycleAdapter(
+  private val navigable: Navigable,
+  private val context: Activity,
+  private val containerRes: Int
+) : DefaultLifecycleObserver {
+
+  override fun onCreate(owner: LifecycleOwner) {
+    navigable.create(context)
```
ryanmoelter, line 19: Won't this call onCreate more times than we want it to? (e.g. every time the activity is recreated)

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
narenkmanoharan: Yeah. I thought about it more and the pros of having distinct events is that onShow/onHide is triggered less often but cons is that you have different events triggered for foreground/background vs navigation and config changes. its a little harder to think about and has caused problems in the past. Also this reduces the complexity and makes us think about it the same way as an activity.

Also i'm gonna start with this and if we see any blocker in the future, we can definitely incorporate it. But for now, this should do. I'm gonna ask the same question in android eng to make sure i'm not missing anything.

PR #117, opened 2020-07-13
Add Navigator for navigating between navigables

narenkmanoharan:
This is a little more flexible than our current navigator as in this navigator can be instantiated inside any screen (without a root screen). Might change that decision in the future but seems fine so far.

Along with the same ability as our past navigator this implementation adds,

- Navigation event to remember and reverse the navigationType so that we don't have to call show and then hide.
- Add the ability to get the path to the root screen for printing the event description

Will be adding more functionality in separate PRs in the future.


magellan-library/src/test/java/com/wealthfront/magellan/navigation/NavigatorTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    assertThat(navigable1.getScreenTraversal()).isEqualTo(listOf(navigable1, navigable2, navigable3))
+  }
+
+  @Test
+  fun getTraversalDescription() {
+    navigator.goTo(navigable1)
+    navigator.goTo(navigable2)
+    navigator.goTo(navigable3)
+
+    assertThat(navigable1.getTraversalDescription())
```
narenkmanoharan, line 143: Will add a androidTest when i introduce a sample to test the traversal of a multi-navigator description.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -0,0 +1,187 @@
...
+  fun replaceAndGo(nextNavigable: Navigable) {
+    replace(nextNavigable, GO)
+  }
+
+  fun replaceAndShow(nextNavigable: Navigable) {
+    replace(nextNavigable, SHOW)
+  }
+
+  private fun replace(nextNavigable: Navigable, navType: NavigationType) {
+    navigate(FORWARD) {
```
ryanmoelter, line 88: nit: this implicitly uses our backstack instead of the argument

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -0,0 +1,187 @@
...
+        if (context != null) {
+          containerView!!.removeView(from)
+          if (from == ghostView) {
+            // Only clear the ghost if it's the same as the view we just removed
+            ghostView = null
+          }
+          currentNavigable!!.transitionFinished()
+          containerView!!.setInterceptTouchEvents(false)
+        }
+      }
```
ryanmoelter, line 136: I think we need a `currentNavigable.transitionStarted()` in here somewhere

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -0,0 +1,187 @@
...
+    containerView = container()
+    currentNavigable?.let {
+      containerView!!.addView(it.view!!)
+    }
+  }
+
+  override fun onDestroy(context: Context) {
+    currentNavigable?.let {
+      lifecycleStateMachine.transitionBetweenLifecycleStates(it, Created(context), Destroyed)
+    }
```
ryanmoelter, line 63: The `currentNavigable` should be attached to the lifecycle already, no?

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -0,0 +1,187 @@
...
+    lifecycleStateMachine.transitionBetweenLifecycleStates(
+      backStack.navigables(),
+      Created(context),
+      Destroyed)
```
ryanmoelter, line 67: When we introduce `navigator.showAbove` or whatever we'll call it, we should also introduce a way to continue to be attached to the lifecycle owner but not receive events past a certain point. For example, the screens in our backstack could still be attached, just not receive any events past `create`, or a screen behind one that's "shown above" it would still be attached, just not receive any events past `start`/`show`

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
narenkmanoharan: fixed

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
narenkmanoharan: Moved it near the transition start.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
narenkmanoharan: Yup. There's a ticket for that.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
narenkmanoharan: I believe thats true. Removed.

PR #120, opened 2020-07-16
Introduce Journey/Flow and a few fixes

narenkmanoharan:
This is a convenience class for a screen which contains a navigator. 

This Pr also has other fixes:

1. Add a magellan_root that we use for setContentView in the root activity. This should allow for easier setup of the library and helps maintain consistency.

2. Rename method name from `transitionBetweenLifecycleStates` to `transition`

3.  Use a stack data structure for the navigator instead of a Deque to limit the API.

4. Add more tests for the linearnavigator.


magellan-library/src/main/java/com/wealthfront/magellan/core/Journey.kt
```diff
@@ -0,0 +1,16 @@
+package com.wealthfront.magellan.core
+
+import android.view.LayoutInflater
+import androidx.viewbinding.ViewBinding
+import com.wealthfront.magellan.ScreenContainer
+import com.wealthfront.magellan.lifecycle.lifecycle
+import com.wealthfront.magellan.navigation.LinearNavigator
+
+abstract class Journey<V : ViewBinding>(
```
narenkmanoharan, line 9: Introducing `Journey` 🚣

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -2,46 +2,52 @@ package com.wealthfront.magellan.lifecycle
...
 internal class ActivityLifecycleAdapter(
   private val navigable: Navigable,
-  private val context: Activity,
-  private val containerRes: Int
+  private val context: Activity
 ) : DefaultLifecycleObserver {
 
-  override fun onStart(owner: LifecycleOwner) {
+  override fun onCreate(owner: ActivityLifecycleOwner) {
+    if (navigable is MagellanLifecycleOwner && navigable.currentState == LifecycleState.Destroyed) {
```
narenkmanoharan, line 18: Makes sure that onCreate is only called after all the navigables are destroyed (application is destroyed)

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
```diff
@@ -16,12 +16,12 @@ internal class LifecycleRegistry : LifecycleAware {
       val oldState = field
       field = newState
       if (listeners.isNotEmpty()) {
-        lifecycleStateMachine.transitionBetweenLifecycleStates(listeners, oldState, newState)
+        lifecycleStateMachine.transition(listeners, oldState, newState)
```
narenkmanoharan, line 19: rename the method.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -16,37 +16,31 @@ import com.wealthfront.magellan.lifecycle.LifecycleState.Created
...
 
-  private val lifecycleStateMachine = LifecycleStateMachine()
+  private var containerView: ScreenContainer? = null
 
   @VisibleForTesting
-  override val backStack: LinkedList<NavigationEvent> = LinkedList()
-
-  private var ghostView: View? = null
-  private var containerView: ScreenContainer? = null
+  override val backStack: Stack<NavigationEvent> = Stack()
```
narenkmanoharan, line 30: Using a stack now so we need to do the empty check but also restrict the api a lot which is good here.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -55,13 +49,10 @@ class LinearNavigator(
...
-    currentNavigable?.let {
-      lifecycleStateMachine.transitionBetweenLifecycleStates(it, Created(context), Destroyed)
+    backStack.navigables().forEach {
+      removeFromLifecycle(it, detachedState = Destroyed)
     }
-    lifecycleStateMachine.transitionBetweenLifecycleStates(
-      backStack.navigables(),
-      Created(context),
-      Destroyed)
+    backStack.clear()
```
narenkmanoharan, line 55: remove the navigables from the lifecycle and clear the backstack whenever the application onDestroy is called.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
ryanmoelter: Probably an edge case we don't care about, but this won't call `onCreate` if the root navigable isn't a `LifecycleOwner`.

I can't think of a time where that would be the case though haha

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigator.kt
```diff
@@ -16,37 +16,31 @@ import com.wealthfront.magellan.lifecycle.LifecycleState.Created
...
   private val container: () -> ScreenContainer
 ) : Navigator, LifecycleAwareComponent() {
 
-  private val lifecycleStateMachine = LifecycleStateMachine()
+  private var containerView: ScreenContainer? = null
 
   @VisibleForTesting
-  override val backStack: LinkedList<NavigationEvent> = LinkedList()
-
-  private var ghostView: View? = null
```
ryanmoelter, line 34: Nice that we don't need a `ghostView` anymore

magellan-library/src/main/java/com/wealthfront/magellan/core/Journey.kt
```diff
@@ -0,0 +1,16 @@
...
+  createBinding: (LayoutInflater) -> V,
+  container: V.() -> ScreenContainer
+) : Screen<V>(createBinding) {
```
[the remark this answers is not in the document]
narenkmanoharan: Yup!

PR #125, opened 2020-07-17
Move common code into magellan-library

narenkmanoharan:
This PR moves common code like the screencontainer and transitions into magellan-common so that we can share this across legacy and library.


magellan-library/build.gradle
```diff
@@ -42,8 +42,7 @@ android {
 }
 
 dependencies {
-
-  implementation project(':magellan-legacy')
```
narenkmanoharan, line 46: no more dependency on legacy ✨

ryanmoelter: Why not move these into Magellan library and have legacy depend on that? That way we can avoid creating another module, and we're going to need that dependency when we make the compatibility navigator in legacy

narenkmanoharan: I was thinking we would make the compatibility navigator in magellan-support. But i suppose this works too.

narenkmanoharan: Done

PR #128, opened 2020-07-20
Add NavigationTraverser to get GlobalBackStack

narenkmanoharan:
This will allow us to get the state of the global backstack at any point in time. We do require the root at any point in time, so this class expects the rootscreen. With depedency injection it should be much easier. 

Bonus: Use Androidx application for context in tests instead of deprecated Robolectric application


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -0,0 +1,34 @@
...
+      val nav = navEvent.navigable
+      backStackList.add(nav)
+      if (nav is Journey<*>) {
+        exploreBackStack(nav, backStackList)
+      }
+    }
+  }
+
+  private fun getBackStackForJourney(journey: Journey<*>): List<NavigationEvent> {
+    return journey.children.mapNotNull { it as? Navigator }.first().backStack
```
narenkmanoharan, line 32: will add a lint rule to ship with our library to make sure that each journey has just one navigator and also i suppose we need to check that nothing outside of a journey has a navigator as well.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -0,0 +1,34 @@
+package com.wealthfront.magellan.navigation
+
+import com.wealthfront.magellan.core.Journey
+import com.wealthfront.magellan.core.Navigable
+
+class NavigationTraverser(private val root: Journey<*>) {
```
ryanmoelter, line 6: Why don't we use `LifecycleOwner` here?

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
narenkmanoharan: Talked offline. We decided that if we wanted to do nested/child navigation, we could always make a child journey. And we _do_ want to enforce the condition of each journey having just one backstack/navigation.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  fun getGlobalBackStack(): List<Navigable> {
```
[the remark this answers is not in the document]
narenkmanoharan: Talked offline. Agreed on this to make it return a tree which can be visualized later.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
narenkmanoharan: This doesn't need to be a `LifecycleOwner `

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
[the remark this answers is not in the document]
narenkmanoharan: Will improve the visualization of the backstack in a separate PR..

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
ryanmoelter: Yeah I have no idea what I was thinking haha

PR #129, opened 2020-07-20
Add NavigationPropagator to propagate navigation events

narenkmanoharan:
This allows us to listen to onShow/onHide events of our navigables. And in general, just an onNavigate event to reset the menu, etc..

### DIFF
--- a/magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
+++ b/magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
@@ -1,16 +1,44 @@
 package com.wealthfront.magellan.sample
 
 import android.content.Context
+import android.util.Log
 import com.wealthfront.magellan.core.Journey
+import com.wealthfront.magellan.core.Navigable
+import com.wealthfront.magellan.navigation.NavigationPropagator.addNavigableListener
+import com.wealthfront.magellan.navigation.NavigationPropagator.removeNavigableListener
+import com.wealthfront.magellan.navigation.NavigationTraverser
+import com.wealthfront.magellan.navigation.NavigableListener
 import com.wealthfront.magellan.sample.databinding.ExpeditionBinding
 
-class Expedition : Journey<ExpeditionBinding>(ExpeditionBinding::inflate, ExpeditionBinding::container) {
+class Expedition : Journey<ExpeditionBinding>(
+  ExpeditionBinding::inflate,
+  ExpeditionBinding::container
+), NavigableListener {
+
+  private val navigationTraverser = NavigationTraverser(this)
 
   override fun onCreate(context: Context) {
+    addNavigableListener(this)
     navigator.goTo(FirstJourney(::goToSecondJourney))
+  }
+
+  override fun onDestroy(context: Context) {
+    removeNavigableListener(this)
   }
 
   private fun goToSecondJourney() {
     navigator.show(SecondJourney())
+  }
+
+  override fun onNavigableShown(navigable: Navigable) {
+    Log.i(this::class.java.simpleName, "Shown: ${navigable.javaClass.simpleName}")
+  }
+
+  override fun onNavigableHidden(navigable: Navigable) {
+    Log.i(this::class.java.simpleName, "Hidden: ${navigable.javaClass.simpleName}")
+  }
+
+  override fun onNavigate() {
+    navigationTraverser.logGlobalBackStack()
   }
 }
--- a/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationPropagator.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationPropagator.kt
@@ -1,0 +1,34 @@
+package com.wealthfront.magellan.navigation
+
+import com.wealthfront.magellan.core.Navigable
+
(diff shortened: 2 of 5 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -16,6 +17,14 @@ class NavigationTraverser(private val root: Journey<*>) {
     return globalBackStack.filter { it.view != null }
   }
 
+  fun logGlobalBackStack() {
```
narenkmanoharan, line 20: Just a helper for logging global navigation events.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationPropagator.kt
```diff
@@ -0,0 +1,34 @@
...
+    listeners = listeners - navigableListener
+  }
+
+  fun onNavigate() {
+    listeners.forEach {
+      it.onNavigate()
+    }
+  }
+
+  fun hideCurrentNavigable(currentNavigable: Navigable) {
```
narenkmanoharan, line 23: will add a lint rule to ensure that `removeNavigableListener` is always called in the same class where its added.

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
```diff
@@ -1,16 +1,44 @@
...
+import com.wealthfront.magellan.navigation.NavigationPropagator.removeNavigableListener
+import com.wealthfront.magellan.navigation.NavigationTraverser
+import com.wealthfront.magellan.navigation.NavigableListener
 import com.wealthfront.magellan.sample.databinding.ExpeditionBinding
 
-class Expedition : Journey<ExpeditionBinding>(ExpeditionBinding::inflate, ExpeditionBinding::container) {
+class Expedition : Journey<ExpeditionBinding>(
+  ExpeditionBinding::inflate,
+  ExpeditionBinding::container
+), NavigableListener {
```
ryanmoelter, line 16: Nit: I think we should create an anonymous implementation of `NavigableListener` instead. Especially in a demo app, it's hard to tell what methods are from which interface when there's one class implementing them all

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationPropagator.kt
ryanmoelter: Could we add it as a lifecycle-aware component instead?

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -16,6 +17,14 @@ class NavigationTraverser(private val root: Journey<*>) {
...
+    val navHistory = getGlobalBackStack().joinToString(" -> ") {
+      val type = if (it is Journey<*>) { "(Journey) " } else { "(Step) " }
```
ryanmoelter, line 22: Minor: I would like to avoid having references to our concrete implementations (`Journey` and `Step`) as much as possible. We should consider checking for the existence of a navigator instead, but if that's too much work, then nevermind

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
ryanmoelter: On second thought, we should absolutely have this as a separate class and attach it to the lifecycle via `by lifecycle()`

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationPropagator.kt
narenkmanoharan: done.

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
narenkmanoharan: done

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
narenkmanoharan: Done.

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
```diff
@@ -2,9 +2,18 @@ package com.wealthfront.magellan.sample
...
+import com.wealthfront.magellan.navigation.NavigationTraverser
 import com.wealthfront.magellan.sample.databinding.ExpeditionBinding
 
-class Expedition : Journey<ExpeditionBinding>(ExpeditionBinding::inflate, ExpeditionBinding::container) {
+class Expedition : Journey<ExpeditionBinding>(
+  ExpeditionBinding::inflate,
+  ExpeditionBinding::container
+) {
+
+  private val navigationTraverser = NavigationTraverser(this)
```
narenkmanoharan, line 15: Using injection in upcoming PRs

PR #130, opened 2020-07-20
Add MenuProvider to sample for how menu's will work with 2.0

narenkmanoharan:
Since menus need to be manipulated outside the journey/step's lifecycle we need a longer living object such as the MenuProvider. Its' lifecycle is tied to the activity.

**Bonus:** Set up dependency injection for the sample project.


magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -46,8 +47,9 @@ internal class ActivityLifecycleAdapter(
 }
 
 fun ComponentActivity.setContentScreen(
-  navigable: Navigable
+  navigable: Navigable,
+  @LayoutRes root: Int = R.layout.magellan_root
```
narenkmanoharan, line 51: Allows users to set custom layouts if necessary.

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt
```diff
@@ -1,21 +1,61 @@
...
 
   override fun onBackPressed() {
-    if (!expeditionProvider.expedition.backPressed()) {
+    if (!expedition.backPressed()) {
       super.onBackPressed()
     }
   }
+
+  override fun onDestroy() {
+    menuProvider.clearMenu()
```
ryanmoelter, line 48: Do we want to make menuProvider a `LifecycleAware` component and do this clearing in `onHide`?

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt
```diff
@@ -1,21 +1,61 @@
...
+  override fun onCreateOptionsMenu(menu: Menu?): Boolean {
+    menuInflater.inflate(R.menu.menu, menu)
+    menuProvider.setMenu(menu)
+    return super.onCreateOptionsMenu(menu)
+  }
+
+  override fun onPrepareOptionsMenu(menu: Menu?): Boolean {
+    menuProvider.setMenu(menu)
+    return super.onPrepareOptionsMenu(menu)
   }
```
ryanmoelter, line 39: What's the difference between these?

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt
```diff
@@ -1,21 +1,61 @@
...
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    setContentScreen(expeditionProvider.expedition)
+    appComponent.inject(this)
+    setContentScreen(expedition)
+    initActionBar()
+  }
+
+  private fun initActionBar() {
+    val actionBar = super.getSupportActionBar()!!
```
ryanmoelter, line 25: Do we want to use a `Toolbar` here, since that's google's recommended way of doing this?

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/menu/MenuProvider.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  fun clearMenu() {
+    menu = null
+  }
+
+  fun findItem(item: Int): MenuItem {
+    return menu!!.findItem(item)
+  }
+
+  fun updateMenu() {
```
ryanmoelter, line 26: nit: we should make this name more descriptive, like `clearMenu` or `makeAllItemsInvisible`

magellan-journey-sample/src/main/res/values/strings.xml
```diff
@@ -7,4 +7,5 @@
   <string name="detail_memo_text">This is detailed memo of your journey!!</string>
   <string name="learn_more_text">Here you will learn more about the expedition!</string>
   <string name="icon">icon</string>
+  <string name="activities">Activities</string>
```
ryanmoelter, line 10: Nit: we probably shouldn't call this activities, since it conflicts with android's `Activity`. Maybe `Notifications` or `Alerts` or something?

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt
narenkmanoharan: Done

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt
narenkmanoharan:
```
onCreateOptionsMenu() is called once, the first time the options menu is displayed
onPrepareOptionsMenu() is called every time the menu opens
``

File: magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt (hunk 4 above)
Comment (narenkmanoharan) [reply magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/MainActivity.kt:25]: uh. Im just gonna leave this as is since its just a sample..

File: magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/menu/MenuProvider.kt (hunk 5 above)
Comment (narenkmanoharan) [reply magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/menu/MenuProvider.kt:26]: done

File: magellan-library/src/main/java/com/wealthfront/magellan/core/Navigable.kt [hunk 7]
```diff
@@ -1,12 +1,15 @@
...
 
+import android.view.Menu
 import android.view.View
 import com.wealthfront.magellan.lifecycle.LifecycleAware
 
 interface Navigable : LifecycleAware {
 
   val view: View?
 
+  fun onUpdateMenu(menu: Menu) {}
```
ryanmoelter, line 11: Hmm, do we actually want to do this? Will it propagate it to its children? This doesn't feel like a concern of a Navigable.

The MenuProvider isn't really part of the core library, as it's structured today. So maybe instead of adding this to the `Navigable` interface, we should make a separate interface that Steps can implement when they want to update the menu.

I'm not sure though. Is that too much work? But it also feels weird to add onUpdateMenu to Navigable and not have it as part of the lifecycle or anything.

magellan-journey-sample/src/main/java/com/wealthfront/magellan/sample/menu/MenuProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import android.content.Context
+import android.view.Menu
+import com.wealthfront.magellan.core.Navigable
+import com.wealthfront.magellan.navigation.NavigableListener
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class MenuProvider @Inject constructor() : NavigableListener {
```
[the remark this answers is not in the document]
narenkmanoharan: Oh my bad. Moved it into the library.

magellan-library/src/main/java/com/wealthfront/magellan/core/Navigable.kt
narenkmanoharan: I updated the `MenuComponent` so that it propagates the onUpdateMenu to the navigables, both the journey and the navigable and its children. I think its fine to keep the onUpdateMenu on the Navigable interface for now since all we are saying is that all navigable are gonna be able to control the menu (just like we saw each navigable is gonna be associated with an android view), which i think is valid.

ryanmoelter: That makes sense, but I don't like that it doesn't work out of the box when you attach the expedition to the activity. Imo, we either need to have it completely handled by Magellan (which means we'd probably need to make an official `Expedition` class and make adding more connection points to the `MainActivity` our official connection), or we need to make it completely optional (we don't have an `onUpdateMenu` in the Navigable, and we can make our own API for `MenuProvider`).

Because I would like to avoid creating an actual `Expedition` class, and partially because I think we can come up with a better API than `onUpdateMenu`, I think it should be completely optional. Also, the way this is set up is weirdly close to a lifecycle event without actually using that pattern.

Also, maybe neither of these approaches are correct in the end. We started talking earlier about a way for multiple nested Journeys to control top and bottom chrome without the possibility of it nesting. Maybe this should be a part of that when we build it? If so, I would double down on my opinion that we keep menus out of the core interfaces so it's easier to change later.

narenkmanoharan: Hmm. Let me think about this more. For now, I'll just put this on pause. We can come back to this later.

PR #133, opened 2020-08-01
Update config to run integration tests on travis

narenkmanoharan:


.travis.yml
```diff
@@ -4,18 +4,23 @@ android:
   components:
     - tools
     - platform-tools
-    - build-tools-29.0.2
-    - android-29
+    - build-tools-26.0.2
```
ryanmoelter, line 7: Wait why are we downgrading this so far? Can we still compile with this?

narenkmanoharan: i was just experimenting. Travis CI has zero documentation on this and following the few steps that they have does not work. but finally got it working

PR #134, opened 2020-08-03
Add Dependency injection for magellan sample

narenkmanoharan:
Since I paused the menu component PR, I decoupled the dependency injection i had as a bonus in that PR.

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
@@ -8,7 +8,7 @@ import com.wealthfront.magellan.lifecycle.LifecycleState.Shown
 
 internal class LifecycleRegistry : LifecycleAware {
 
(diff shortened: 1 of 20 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
```diff
@@ -8,7 +8,7 @@ import com.wealthfront.magellan.lifecycle.LifecycleState.Shown
 
 internal class LifecycleRegistry : LifecycleAware {
 
-  var listeners: List<LifecycleAware> = emptyList()
+  var listeners: Set<LifecycleAware> = emptySet()
```
ryanmoelter, line 11: Hmm, I think this changes the ordering of lifecycle event propagation, which could be very bad. I think TreeSet preserves ordering?

narenkmanoharan: fixed. I think TreeSet sorts the items. I need a LinkedHashSet

PR #135, opened 2020-08-05
AP - 2926 - Navigation between magellan 1.0 and 2.0 screens

narenkmanoharan:
- This PR makes our old screen implement navigable and lifecycle aware so we can essentially navigate from new steps into old screens and vice versa
- For navigating from old screen, we will inject the navigator from the "Expedition" class into the screen through dependency injection 
- Also this change makes it so that we don't have to essentially "deprecate" all our current screens since we can navigate to them just like we would with the newer Step/Journey.
- We need a script which migrates the usages on the navigator from the screen to the navigation from the expedition (Looks like we might need a expedition (root journey) class in the library)

_Work left to be done later_

- Support menus probably with the same API
- Support set title for navigables
- Complete support for History Rewriter
  - resetWithRoot


magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/tide/TideDetailsScreen.java
```diff
@@ -32,10 +34,12 @@ public int compare(Observation o1, Observation o2) {
   };
 
   @Inject NoaaApi noaaApi;
+  @Inject @Named("LegacyNavigator") Navigator navigator;
```
narenkmanoharan, line 37: Navigating from one screen to another will now require this injected navigator. You can navigate to both old screens or the new navigables.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAware.kt
```diff
@@ -16,5 +17,7 @@ interface LifecycleAware {
 
   fun destroy(context: Context) {}
 
+  fun onSaveInstanceState(outState: Bundle) {}
```
narenkmanoharan, line 20: Need this to be passed along to all the children if we wanna save the state across application instances.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -37,21 +46,23 @@
  * }
  * </code> </pre>
  */
-public abstract class Screen<V extends ViewGroup & ScreenView> implements BackHandler {
+public abstract class Screen<V extends ViewGroup & ScreenView> extends LifecycleAwareComponent implements Navigable {
```
narenkmanoharan, line 49: old screens are now navigables.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -37,21 +46,23 @@
...
   private V view;
-  private DialogCreator dialogCreator;
-  private boolean dialogIsShowing;
-  private Dialog dialog;
   private SparseArray<Parcelable> viewState;
   private boolean isTransitioning;
   private Queue<TransitionFinishedListener> transitionFinishedListeners = new LinkedList<>();
 
+  public Screen() {
+    attachToLifecycle(dialogComponent, LifecycleState.Destroyed.INSTANCE);
```
narenkmanoharan, line 63: Since we don't have delegates in java..

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/tide/TideDetailsScreen.java
ryanmoelter: Let's think through this real quick.

If you're going from `Screen` -> `Journey`, you have to provide the `Journey` with its navigational dependencies, i.e. the navigation lambdas. Old magellan `Screen`s aren't set up to do this properly, since they aren't encapsulated properly and would instead end up passing a lambda based on the global navigator, but controlled by a `Screen` which is no longer showing. The more proper way of modeling this in new magellan is to have the container in which the `Step` resides providing the navigation lambdas; at the top level, that means the `Expedition` is responsible for providing them. This could be accomplished by a top-level `Router`, or something more specific than a general `Navigator`; something that has a `goToSpecificJourney()` method that provides it with the navigation dependencies it needs.

None of this so far is incompatible with what we have already here.

Now let's look at `Journey` -> `Screen`. In this model, the `Screen` is added to the `Journey`'s backstack like any `Step` would. The problem is that `Screen`s have a very different navigational model than `Step`s; `Screens` do all navigation at the top level, which breaks the expectation that `Journey`s will continue to be `show`n until they themselves navigate away. It also breaks the mental model of `Journey`s by effectively adding a goto statement in the middle of our flow that may or may not ever return.

This mismatch of navigation expectations is what I'm most worried about, and I'm worried that it's impossible to avoid if we make `Screens` implement `Navigable`.

(Oh, also, for the same reason that I think `Screen`s being able to interrupt a `Journey` is dangerous, I think exposing a "LegacyNavigator" is a dangerous pattern if it spreads into our `Journey`s/`Step`s.)

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -199,26 +188,41 @@ protected void onRestore(Bundle savedInstanceState) {}
...
+  protected void onHide(@NotNull Context context) {
+    destroyView();
+  }
+
+  /**
+   * Called when the activity is killed.
+   */
+  @Override
+  @CallSuper
+  protected void onDestroy(@NotNull Context context) { }
```
ryanmoelter, line 225: Do we need these empty override methods or can we just inherit them from `LifecycleAware`?

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -37,21 +46,23 @@
...
-public abstract class Screen<V extends ViewGroup & ScreenView> implements BackHandler {
+public abstract class Screen<V extends ViewGroup & ScreenView> extends LifecycleAwareComponent implements Navigable {
 
   public static final int DEFAULT_ACTION_BAR_COLOR_RES = 0;
   private static final String VIEW_STATE = "com.wealthfront.navigation.Screen.viewState";
 
+  private final DialogComponent dialogComponent = new DialogComponent();
+
   private Activity activity;
-  private Navigator navigator;
```
ryanmoelter, line 46: This is going to break a lot of code without a migration. If we want to do this, we should provide an easy-to-use script that will update all `Screen`s with whatever new pattern we recommend, otherwise this will be painful.

magellan-library/src/main/java/com/wealthfront/magellan/view/DialogComponent.kt
```diff
@@ -1,12 +1,12 @@
 package com.wealthfront.magellan.view
 
+import android.app.Activity
 import android.app.Dialog
 import android.content.Context
+import com.wealthfront.magellan.DialogCreator
 import com.wealthfront.magellan.lifecycle.LifecycleAware
 import javax.inject.Inject
 
-typealias DialogCreator = (Context) -> Dialog
```
ryanmoelter, line 8: We want to use the Java interface for java interop? Could we overload the method so we can use nice Kotlin syntax without a SAM conversion for our new `Step`s?

magellan-legacy/src/main/java/com/wealthfront/magellan/ScreenGroup.java
```diff
@@ -37,9 +37,9 @@ public void addScreens(List<S> screens) {
 
   @Override
   protected void onShow(Context context) {
+    super.onShow(context);
```
ryanmoelter, line 40: Isn't `ScreenGroup` obsolete if `Screen`s are `Navigable`? Should we just make this an empty class?

magellan-legacy/src/test/java/com/wealthfront/magellan/ScreenTest.java
```diff
@@ -45,65 +42,57 @@ public void setUp() {
 
   @Test
   public void recreateView() {
-    View v = screen.recreateView(null, null);
+    View v = screen.recreateView(null);
```
ryanmoelter, line 45: What calls `recreateView` if Screens are Navigable? It's not part of Navigable, which means it's not guaranteed to be called

ryanmoelter: Same with `destroyView` below

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAware.kt
ryanmoelter: Not to mention this requires a separate attachment point to the Activity. Also it's useless if we don't pass the bundle to `onCreate`

magellan-library/src/main/java/com/wealthfront/magellan/view/ViewExtensions.kt
```diff
@@ -3,7 +3,7 @@ package com.wealthfront.magellan.view
 import android.view.View
 import android.view.ViewTreeObserver
 
-internal fun View.whenMeasured(onMeasured: () -> Unit) {
+fun View.whenMeasured(onMeasured: () -> Unit) {
```
ryanmoelter, line 6: We probably shouldn't make this public, I think there's a competing method in ktx. Let's double check that before making this public

magellan-library/src/test/java/com/wealthfront/magellan/view/DialogComponentTest.kt
```diff
@@ -29,16 +29,16 @@ class DialogComponentTest {
...
+    dialogCreator2 = DialogCreator { context -> dialog2 }
   }
 
   @Test
   fun showDialog() {
     `when`(dialog1.isShowing).thenReturn(true)
 
     dialogComponent.create(context)
-    dialogComponent.showDialog { dialogCreator1.invoke(context) }
+    dialogComponent.showDialog(DialogCreator { context -> dialogCreator1.createDialog(context) })
```
ryanmoelter, line 41: Um, should we unwrap this and just pass `dialogCreator1` directly?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAware.kt
narenkmanoharan: Talked offline. Remove onSaveInstanceState since its not that useful here.

magellan-library/src/main/java/com/wealthfront/magellan/view/ViewExtensions.kt
narenkmanoharan: d

magellan-library/src/main/java/com/wealthfront/magellan/navigation/Navigator.kt
```diff
@@ -1,6 +1,25 @@
...
+  fun show(nextNavigable: Navigable)
+
+  fun replaceAndGo(nextNavigable: Navigable)
+
+  fun replaceAndShow(nextNavigable: Navigable)
+
+  fun navigate(
+    direction: Direction,
+    backStackOperation: (Stack<NavigationEvent>) -> Unit
+  )
```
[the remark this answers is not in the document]
narenkmanoharan: Removed from the Navigator interface. Created a LegacyNavigator which inherits from LinearNavigator which has the same ability except that it sets up the navigator in the screen.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -199,26 +188,41 @@ protected void onRestore(Bundle savedInstanceState) {}
...
 
   /**
-   * Called when the Activity is resumed and when the Screen is shown.
+   * Called when the Screen in shown (including on rotation).
    */
-  protected void onResume(Context context) {}
+  @Override
+  @CallSuper
+  protected void onShow(@NotNull Context context) {
+    recreateView(context);
```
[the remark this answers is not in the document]
narenkmanoharan: Created a LegacyViewComponent which does the onShow/onHide setup so we don't have to call super anymore and also handles saving and restoring view state.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -199,26 +188,41 @@ protected void onRestore(Bundle savedInstanceState) {}
   protected void onUpdateMenu(Menu menu) {}
 
   /**
-   * Called when the Activity is resumed and when the Screen is shown.
+   * Called when the Screen in shown (including on rotation).
    */
-  protected void onResume(Context context) {}
+  @Override
+  @CallSuper
+  protected void onShow(@NotNull Context context) {
```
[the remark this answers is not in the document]
narenkmanoharan: onShow's behaviour will be different and that is intentional, it helps model screens and steps/journeys the same way so that its easier to reason about in the future. I'm not too worried since 2.0 is gonna be a breaking change anyways.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: We need these since its protected in the parent class.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: Added back the navigator.

magellan-library/src/main/java/com/wealthfront/magellan/view/DialogComponent.kt
narenkmanoharan: Overloaded the `showDialog` method.

magellan-legacy/src/main/java/com/wealthfront/magellan/ScreenGroup.java
narenkmanoharan: Changed screen to only add screen and attach them to the screengroups lifecycle.

magellan-legacy/src/test/java/com/wealthfront/magellan/ScreenTest.java
narenkmanoharan: Removed these.

magellan-library/src/test/java/com/wealthfront/magellan/view/DialogComponentTest.kt
narenkmanoharan: d

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: Not anymore..

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/tide/TideDetailsScreen.java
narenkmanoharan: Agreed. Removed the injected navigator and exposing a legacy navigator through the screens which can only navigate at the expedition level. Our linear navigator forbids navigating to our old screens.

magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyNavigator.kt
```diff
@@ -0,0 +1,54 @@
...
+
+  private val delegate by lifecycle(NavigationDelegate(container))
+
+  override val backStack: Stack<NavigationEvent>
+    get() = delegate.backStack
+
+  init {
+    delegate.currentNavigableSetup = { navItem ->
+      if (navItem is Screen<*>) {
+        navItem.setNavigator(this)
```
ryanmoelter, line 23: Don't we need to do this when any screen is navigated to?

magellan-library/src/main/java/com/wealthfront/magellan/view/DialogComponent.kt
```diff
@@ -37,9 +43,13 @@ class DialogComponent @Inject constructor() : LifecycleAware {
     this.context = null
   }
 
+  fun getDialog(): Dialog? {
+    return dialog
+  }
```
ryanmoelter, line 48: nit: make the `var dialog` public with a `private set`

magellan-library/src/test/java/com/wealthfront/magellan/view/DialogComponentTest.kt
```diff
@@ -29,16 +29,16 @@ class DialogComponentTest {
...
+    dialogCreator2 = DialogCreator { context -> dialog2 }
   }
 
   @Test
   fun showDialog() {
     `when`(dialog1.isShowing).thenReturn(true)
 
     dialogComponent.create(context)
-    dialogComponent.showDialog { dialogCreator1.invoke(context) }
+    dialogComponent.showDialog { context -> dialogCreator1.createDialog(context) }
```
ryanmoelter, line 41: I think I said this before, but nit: these should probably just pass the dialogCreators directly, e.g. `dialogComponent.showDialog(dialogCreator1)`

magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyViewComponent.java
```diff
@@ -0,0 +1,63 @@
...
+  private final Screen<V> screen;
+  private SparseArray<Parcelable> viewState;
+
+  public LegacyViewComponent(Screen<V> screen) {
+    this.screen = screen;
+  }
+
+  @Override
+  public void create(@NotNull Context context) {
+    screen.setActivity((Activity) context);
```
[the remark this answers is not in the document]
narenkmanoharan: It should. We pass the activity context to the `ActivityLifecycleAdapter` through the setContentScreen. I don't believe we pass the `Application context` anywhere.

magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyViewComponent.java
```diff
@@ -0,0 +1,63 @@
...
+    if (view != null) {
+      viewState = new SparseArray<>();
+      view.saveHierarchyState(viewState);
+    }
+    screen.setView(null);
+  }
+
+  @Override
+  public void destroy(@NotNull Context context) {
+    screen.setActivity(null);
```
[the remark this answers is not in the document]
narenkmanoharan: Good catch.

magellan-legacy/src/test/java/com/wealthfront/magellan/ScreenTest.java
```diff
@@ -44,70 +33,45 @@ public void setUp() {
...
-    screen.recreateView(null, null);
-    Bundle bundle = new Bundle();
-    screen.save(bundle);
-    screen.restore(bundle);
-    screen.recreateView(null, null);
-
-    verify(view).saveHierarchyState(isA(SparseArray.class));
-    verify(view).restoreHierarchyState(sparseArrayCaptor.capture());
-    assertThat(((Bundle) sparseArrayCaptor.getValue().get(42)).getString("key")).isEqualTo("value");
+  public void destroyedActivity() {
```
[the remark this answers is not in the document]
narenkmanoharan: The above test tests the behaviour. Changed to make sure we null the activity in onHide.

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -198,27 +149,44 @@ protected void onRestore(Bundle savedInstanceState) {}
    */
   protected void onUpdateMenu(Menu menu) {}
 
-  /**
-   * Called when the Activity is resumed and when the Screen is shown.
-   */
-  protected void onResume(Context context) {}
+  @Override
+  protected void onCreate(@NotNull Context context) { }
```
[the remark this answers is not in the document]
narenkmanoharan: Mostly for java doc purposes.

magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyNavigator.kt
narenkmanoharan: I set this lambda to be executed when the screen is navigated to in the `NavigationDelegate`

magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyViewComponent.java
```diff
@@ -0,0 +1,63 @@
...
+      view.restoreHierarchyState(viewState);
+    }
+    screen.setView(view);
+  }
+
+  @Override
+  public void resume(@NotNull Context context) { }
+
+  @Override
+  public void pause(@NotNull Context context) { }
```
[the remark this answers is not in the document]
narenkmanoharan: Removed

magellan-legacy/src/test/java/com/wealthfront/magellan/LegacyViewComponentTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Before
+  public void setUp() {
+    initMocks(this);
+    screen = new DummyScreen(view);
+    legacyViewComponent = new LegacyViewComponent(screen);
+  }
+
+  @Test
+  public void viewComponentLifecycleUntilDestroy() {
```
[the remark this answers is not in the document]
narenkmanoharan: Added more assertions in the same test and a new case for testing config changes.

magellan-legacy/src/test/java/com/wealthfront/magellan/ScreenGroupTest.java
```diff
@@ -48,6 +44,8 @@ protected ViewGroup createView(Context context) {
 
   @Test
   public void addScreen() {
+    screenGroup.resume(context);
```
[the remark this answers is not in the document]
narenkmanoharan: Yup.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAware.java
```diff
@@ -0,0 +1,22 @@
...
+  default void pause(Context context) {}
+
+  default void hide(Context context) {}
+
+  default void destroy(Context context) {}
+
+  default boolean backPressed() {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
narenkmanoharan: It does mean that i have to set the kotlinOptions to 1.8 for the library. I dont think it should be a problem just calling it out.

magellan-library/src/main/java/com/wealthfront/magellan/view/DialogComponent.kt
narenkmanoharan: d

magellan-library/src/test/java/com/wealthfront/magellan/view/DialogComponentTest.kt
narenkmanoharan: d

magellan-library/build.gradle
```diff
@@ -23,6 +23,11 @@ android {
     setSourceCompatibility(JavaVersion.VERSION_1_8)
     setTargetCompatibility(JavaVersion.VERSION_1_8)
   }
+
+  kotlinOptions {
```
narenkmanoharan, line 27: Need this for supporting `@JvmDefault`

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -71,75 +77,22 @@ public final Activity getActivity() {
...
-  void transitionStarted() {
+  @Override
+  public void transitionStarted() {
     isTransitioning = true;
     transitionFinishedListeners.clear();
   }
 
-  void transitionFinished() {
+  @Override
+  public void transitionFinished() {
```
ryanmoelter, line 95: If we're making this public, we should probably make it final too. The existing API for this is to use `whenTransitionFinished { }`

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -243,49 +217,31 @@ protected final void setTitle(CharSequence title) {
...
   }
 
-  @VisibleForTesting
   public final void setActivity(Activity activity) {
     this.activity = activity;
   }
 
-  @VisibleForTesting
-  public final void setNavigator(Navigator navigator) {
+  public void setNavigator(@NotNull LegacyNavigator navigator) {
```
ryanmoelter, line 241: `internal`? Or `@VisibleForTesting`?

magellan-library/build.gradle
ryanmoelter: We should include this in the breaking changes list, though I doubt it'll affect many people

narenkmanoharan: Yup!

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: d

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: internal does exist in java unfortunately. Changed to package-private.

PR #137, opened 2020-08-10
Add RxDisposer and NavigableScope for RxJava and Coroutines

narenkmanoharan:


magellan-library/src/main/java/com/wealthfront/magellan/coroutines/NavigableScope.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.content.Context
+import com.wealthfront.magellan.lifecycle.LifecycleAware
+import javax.inject.Inject
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+
+class NavigableScope @Inject constructor() : LifecycleAware, CoroutineScope {
```
narenkmanoharan, line 12: Pretty opinionated but I think we should ship this with the core library.

magellan-library/src/main/java/com/wealthfront/magellan/coroutines/NavigableScope.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  private var job = SupervisorJob().apply { cancel(CancellationException("Not created yet")) }
```
narenkmanoharan, line 14: Not entirely sure how to test this. @ryanmoelter any thoughts?

narenkmanoharan: Looks like i need `runBlockingTest`. Pretty interesting.

PR #139, opened 2020-08-11
Add mockable support for mocking library classes

narenkmanoharan:
Add better support for for testing


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationEvent.kt
```diff
@@ -1,8 +1,10 @@
 package com.wealthfront.magellan.navigation
 
+import com.wealthfront.magellan.Mockable
 import com.wealthfront.magellan.NavigationType
 import java.util.Stack
 
+@Mockable
 data class NavigationEvent(
```
ryanmoelter, line 8: I don't think we need to make this `@Mockable`, can you even extend data classes?

PR #141, opened 2020-08-17
Add support for menu and title in both magellan 1.0 and 2.0

narenkmanoharan:
This PR adds the support for title and menu's on both Screens and Steps. I came to the conclusion to keep all the setup being taken care of by the library through the navigation delegate. This way we have the same API and keep compatibility with existing screen (We can change the improve on the API later, if necessary. For now, I think this should be fine).


magellan-library/src/main/java/com/wealthfront/magellan/view/ActionBarModifier.java
```diff
@@ -0,0 +1,38 @@
+package com.wealthfront.magellan.view;
+
+import android.view.Menu;
+import org.jetbrains.annotations.NotNull;
+import androidx.annotation.ColorRes;
+
+public interface ActionBarModifier {
```
narenkmanoharan, line 7: This needs to be in Java since this is a public interface that consumers of the library would need to implement. If we use @JvmDefault here instead, consumers would need to enable `-Xjvm-default` option or else would run into a compile time error. So instead i use a java interface here to maintain compatibility.

```
Inheritance from an interface with '@JvmDefault' members is only allowed with -Xjvm-default option
```

magellan-library/src/main/java/com/wealthfront/magellan/core/Step.kt
```diff
@@ -8,21 +9,28 @@ import android.view.View
...
+  final override var activity: Activity? = null
+    private set
 
   var viewBinding: V? by lifecycleWithContext { createBinding.invoke(LayoutInflater.from(it)) }
     @VisibleForTesting set
 
   final override var view: View? by lifecycleWithContext { viewBinding!!.root }
     @VisibleForTesting set
 
+  override fun onCreate(context: Context) {
```
ryanmoelter, line 28: This will get overridden by inheritors of this class. We should probably create an activity LifecycleComponent instead for this reference

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -160,4 +182,39 @@ class NavigationDelegate(
...
+  private fun updateMenu(menu: Menu?, navItem: NavigableCompat? = null) {
+    // Need to post to avoid animation bug on disappearing menu
+    Handler(Looper.getMainLooper()).post {
+      menu?.let {
+        for (i in 0 until menu.size()) {
+          menu.getItem(i).isVisible = false
+          (rootNavigable as? ActionBarModifier)?.onUpdateMenu(menu)
+          rootNavigable.childNavigables().filterIsInstance(ActionBarModifier::class.java).forEach { it.onUpdateMenu(menu) }
+          (navItem as? ActionBarModifier)?.onUpdateMenu(menu)
+          navItem?.childNavigables()?.filterIsInstance(ActionBarModifier::class.java)?.forEach { it.onUpdateMenu(menu) }
```
ryanmoelter, line 204: Wait, we want to do this for every menu item? I think we want to move this out of the for loop

magellan-library/src/main/java/com/wealthfront/magellan/navigation/Navigator.kt
```diff
@@ -2,5 +2,7 @@ package com.wealthfront.magellan.navigation
 
 interface Navigator {
 
+  val journey: NavigableCompat
```
ryanmoelter, line 5: Is this supposed to be the rootNavigable? Why does every navigator need a reference to a Journey?

magellan-library/src/test/java/com/wealthfront/magellan/navigation/LinearNavigatorTest.kt
```diff
@@ -9,43 +11,71 @@ import com.wealthfront.magellan.ScreenContainer
...
+    linearNavigator.menu = menu
+
+    verify(journey1 as ActionBarModifier).onUpdateMenu(menu)
+
     linearNavigator.goTo(step1)
 
+    verify(step1).setTitle(anyString())
+    verify(journey1 as ActionBarModifier, times(2)).onUpdateMenu(menu)
+    verify(step1 as ActionBarModifier).onUpdateMenu(menu)
+    verify(context as ActionBarConfigListener).onNavigate(any())
```
ryanmoelter, line 78: Nit, not blocking: could we test this by having the journey/step modify the menu, rather than using spies?

magellan-library/src/test/java/com/wealthfront/magellan/navigation/LinearNavigatorTest.kt
```diff
@@ -121,17 +153,17 @@ class LinearNavigatorTest {
...
       it.push(NavigationEvent(step2, GO))
       it.push(NavigationEvent(journey1, SHOW))
     }
 
-    linearNavigator.resume(getApplicationContext())
     linearNavigator.goBack()
     linearNavigator.goBack()
-    linearNavigator.pause(getApplicationContext())
-    linearNavigator.destroy(getApplicationContext())
+    linearNavigator.destroy(context)
```
ryanmoelter, line 166: Does the navigator handle state transitions from e.g. `resumed` to `destroyed`? Does it transition its children through the proper lifecycle events?

magellan-sample-advanced/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
 
   defaultConfig {
     applicationId "com.wealthfront.magellan.sample"
-    minSdkVersion Versions.minSdkVersion
+    minSdkVersion 21
```
ryanmoelter, line 11: This isn't in Versions anymore?

ryanmoelter (changes requested):
Lots of questions because I'm rusty with this stuff after two weeks haha

magellan-library/src/main/java/com/wealthfront/magellan/core/Step.kt
narenkmanoharan: Oops. Missed that. good catch!

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
narenkmanoharan: Moved out of the loop.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/Navigator.kt
narenkmanoharan: Each navigator is now associated with its respective journey. This way the navigation delegate has access to the journey and can call the onUpdateMenu in the order that we want.

magellan-library/src/test/java/com/wealthfront/magellan/navigation/LinearNavigatorTest.kt
narenkmanoharan: Since the linearNavigator is a lifecycleawarecomponent, that should be taken care of by the lifecycle registry.

magellan-sample-advanced/build.gradle
narenkmanoharan: fixed.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleShownWithContext.kt
```diff
@@ -0,0 +1,41 @@
...
+  var data: V? = null
+    protected set
+
+  override fun create(context: Context) {
+    data = supplier(context)
+  }
+
+  override fun hide(context: Context) {
+    data = null
+  }
```
ryanmoelter, line 34: What happens when it's `SHOWN` for a second time? This isn't symmetrical and will cause the `data` to be null after the first `hide` call.

For simplicity, should we just attach/detach the activity and views at the same time? That is, have a single `lifecycleWithContext` that attaches/detaches when `SHOWN`. We don't have a stable `Context` for the whole time we're `CREATED` anyway, so the current implementation is a bit tricky at best to use correctly.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableListener.kt
```diff
@@ -5,12 +5,13 @@ import com.wealthfront.magellan.lifecycle.LifecycleAware
 
 interface NavigableListener : LifecycleAware {
 
-  fun onNavigate() {}
-
   fun onNavigableShown(navigable: NavigableCompat) {}
 
   fun onNavigableHidden(navigable: NavigableCompat) {}
 
+  @JvmDefault
```
ryanmoelter, line 12: Why is this the only one marked with `@JvmDefault`?

ryanmoelter (approved):
We need to fix the `lifecycleCreatedWithContext` method, but other than that, LGTM!

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableListener.kt
narenkmanoharan: I supposed we would always want to implement `onNavigableShown ` and `onNavigableHidden `. But i suppose not.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleShownWithContext.kt
narenkmanoharan: done. Consolidate to just use show/hide.

PR #143, opened 2020-09-03
Add lint rule for preventing usage of activity instance in NavigableCompat

narenkmanoharan:


magellan-lint/src/main/java/com/wealthfront/magellan/AvoidUsingActivity.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.getContainingUClass
+
+internal val AVOID_USING_ACTIVITY = Issue.create(
+  id = AvoidUsingActivity::class.simpleName!!,
+  briefDescription = "Use the context provided in the lifecycle objects instead.",
+  explanation = "Activity is available here only for use to set the title through the navigable compat.",
```
ryanmoelter, line 17: Not: probably don't need to mention Navigable compat here

ryanmoelter (approved):

magellan-lint/src/main/java/com/wealthfront/magellan/AvoidUsingActivity.kt
narenkmanoharan: d

PR #144, opened 2020-09-16
Add appropriate nullability annotations

narenkmanoharan:


magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
```diff
@@ -42,8 +41,8 @@
 
   private final DialogComponent dialogComponent = new DialogComponent();
 
-  private Activity activity;
-  private V view;
+  private @Nullable Activity activity;
+  private @Nullable  V view;
```
ryanmoelter, line 45: nit: space

ryanmoelter (approved):
LGTM! (We may want to set up ktLint to run for PRs in this repo though)

magellan-legacy/src/main/java/com/wealthfront/magellan/Screen.java
narenkmanoharan: d

PR #147, opened 2020-10-07
Transitions Api Update

narenkmanoharan:
With this PR, we will be ridding of navigation type and instead we will only use transitions to model our navigation animations.

Our default goto -> DefaultTransition
Show -> ShowTransition

This PR also makes it so that we automatically reverse the transition without having to manually specify `hide` when we call `show` on the screen. For this, I use the navigation event where we specify the transition that needs to be played when navigating to the screen. This is also required when we are history re-writing now, allowing for greater control of the animations (where we had no control before).

Also I tried implementing the animations using `Blend`, but there were two issues, 1. We need to update our min api to 21 if we wanna do that, 2. I ran into some issues with the translation that caused some weird issue and did not want to spend a lot of time debugging it so keeping the existing animations for now.

Also: Bonus [Update to kotlin 1.4.0](https://github.com/wealthfront/magellan/pull/148)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -151,7 +151,9 @@ class NavigationDelegate(
     callOnNavigate(currentNavigable!!)
     when (currentState) {
       is LifecycleState.Shown, is LifecycleState.Resumed -> {
-        containerView!!.addView(currentNavigable!!.view!!)
+        containerView!!.addView(
+          currentNavigable!!.view!!,
+          direction.indexToAddView(containerView!!))
```
narenkmanoharan, line 156: This was a bug which caused a random glitch while i was testing the transition. Fixed now.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -73,32 +72,32 @@ class NavigationDelegate(
...
-  fun goTo(nextNavigableCompat: NavigableCompat) {
-    navigateTo(nextNavigableCompat, GO)
+  fun goTo(nextNavigableCompat: NavigableCompat, overrideTransition: Transition? = null) {
+    navigateTo(nextNavigableCompat, overrideTransition ?: DefaultTransition())
   }
 
-  fun show(nextNavigableCompat: NavigableCompat) {
-    navigateTo(nextNavigableCompat, SHOW)
+  fun show(nextNavigableCompat: NavigableCompat, overrideTransition: Transition? = null) {
+    navigateTo(nextNavigableCompat, overrideTransition ?: ShowTransition())
```
ryanmoelter, line 80: If we're making this such that `show` is just a transition, we probably should remove `navigator.show()` (while keeping `legacyNavigator.show()` for compatibility)

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -73,32 +72,32 @@ class NavigationDelegate(
...
   }
 
-  fun replaceAndShow(nextNavigableCompat: NavigableCompat) {
-    replace(nextNavigableCompat, SHOW)
+  fun replaceAndShow(nextNavigableCompat: NavigableCompat, overrideTransition: Transition? = null) {
+    replace(nextNavigableCompat, overrideTransition ?: ShowTransition())
   }
 
-  private fun replace(nextNavigableCompat: NavigableCompat, navType: NavigationType) {
+  private fun replace(nextNavigableCompat: NavigableCompat, overrideTransition: Transition? = null) {
```
ryanmoelter, line 91: This is the same as `replaceAndGo` now, right? Could we remove the redundant one? (or better yet, remove `replaceAndGo` and `replaceAndShow` altogether?)

magellan-library/src/main/java/com/wealthfront/magellan/transitions/Transition.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Define a transition (animation) between two screens. By default, transitions are implemented by
+ * [DefaultTransition]. You can either set a different default one when building your Navigator, using
+ * [Navigator.Builder.transition],
+ * or override the next transition by using
+ * [Navigator.overrideTransition].
+ *
+ *
+ * You can find more Transitions implemented in [com.wealthfront.magellan.transitions].
+ */
+interface Transition {
```
ryanmoelter, line 16: Since we're making a breaking change to this interface anyway, how do you feel about renaming this to a `StepTransition` or `MagellanTransition` so it doesn't conflict with android's `Transition` class?

magellan-sample-advanced/src/main/res/layout/dog_breed.xml
```diff
@@ -4,6 +4,7 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="vertical"
+    android:background="@android:color/white"
```
ryanmoelter, line 7: For GPU overdraw reasons, we should encourage users to keep their `Step`'s views transparent. Could we automatically add an `?colorBackground`-colored background to `ShowTransition` and other transitions that overlap?

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -73,32 +72,32 @@ class NavigationDelegate(
...
     navigate(FORWARD) { backStack ->
       backStack.pop()
-      backStack.push(NavigationEvent(nextNavigableCompat, navType))
+      backStack.push(NavigationEvent(nextNavigableCompat, overrideTransition ?: ShowTransition()))
     }
   }
 
-  private fun navigateTo(nextNavigableCompat: NavigableCompat, navType: NavigationType) {
+  private fun navigateTo(nextNavigableCompat: NavigableCompat, overrideTransition: Transition? = null) {
```
ryanmoelter, line 98: Same as above, this is now equivalent to `goTo()` and close to `show()`

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,9 +1,9 @@
 object Versions {
   const val compileSdkVersion = 30
-  const val minSdkVersion = 21
+  const val minSdkVersion = 15
```
ryanmoelter, line 3: I think I made blend `21` min because of elevation. Maybe I can add a compat version that goes back further... We'll see 🙃

ryanmoelter (changes requested):
I love this change, but we should take some time to deduplicate the methods on our new `LinearNavigator`, since a lot of them are the same now. We may also want to emulate the old `NavType` api in the `LegacyNavigator` for easier migration

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
narenkmanoharan: done

magellan-sample-advanced/src/main/res/layout/dog_breed.xml
narenkmanoharan: Not sure about this.

magellan-library/src/main/java/com/wealthfront/magellan/transitions/Transition.kt
narenkmanoharan: Done

magellan-library/src/main/java/com/wealthfront/magellan/transitions/DefaultTransition.kt
```diff
@@ -0,0 +1,39 @@
...
+    to: View,
+    direction: Direction
+  ): AnimatorSet {
+    val sign = direction.sign()
+    val axis: Property<View, Float> = View.TRANSLATION_X
+    val toTranslation = sign * to.width
+    val set = AnimatorSet()
+    if (from != null) {
+      val fromTranslation = sign * -from.width
+      set.play(ObjectAnimator.ofFloat(from, axis, 0f, fromTranslation.toFloat()))
```
ryanmoelter, line 34: Do we set the interpolator somewhere and I'm just missing it?

magellan-library/src/main/java/com/wealthfront/magellan/transitions/MagellanTransition.kt
```diff
@@ -0,0 +1,38 @@
...
+  fun animate(
+    from: View?,
+    to: View,
+    direction: Direction,
+    callback: Callback
+  )
+
+  interface Callback {
+    fun onAnimationEnd()
+  }
```
ryanmoelter, line 37: Could this just be a `() -> Unit` for simplicity?

magellan-library/src/test/java/com/wealthfront/magellan/transitions/CrossfadeTransitionTest.java
```diff
@@ -37,33 +34,18 @@ public void setUp() {
...
 
-  @Test
-  public void animateGoTo() {
-    checkAnimate(GO, FORWARD);
-  }
-
-  @Test
-  public void animateGoBack() {
-    checkAnimate(GO, BACKWARD);
-  }
```
ryanmoelter, line 48: Yessss delete all the code. Simple transitions! 🎉

ryanmoelter (approved):
Just a few thoughts, nothing major

magellan-library/src/main/java/com/wealthfront/magellan/transitions/DefaultTransition.kt
narenkmanoharan: Talked offline. We will do this in a separate PR.

magellan-library/build.gradle
```diff
@@ -62,7 +62,8 @@ dependencies {
   testImplementation Dependencies.robolectric
   testImplementation Dependencies.coroutinesTest
 
-  lintPublish project(':magellan-lint')
+// Bug in AGP: Follow this issue - https://issuetracker.google.com/issues/141840950
+// lintPublish project(':magellan-lint')
```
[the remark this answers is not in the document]
narenkmanoharan: Unfortunately, yeah... I've left a comment on the issues.. Lets see what happens.

magellan-library/src/main/java/com/wealthfront/magellan/transitions/MagellanTransition.kt
narenkmanoharan: done

PR #148, opened 2020-10-07
Update to Kotlin 1.4.0

narenkmanoharan:


magellan-lint/src/main/java/com/wealthfront/magellan/AvoidUsingActivity.kt
```diff
@@ -34,13 +34,13 @@ internal class AvoidUsingActivity : Detector(), Detector.UastScanner {
 
     override fun visitCallExpression(node: UCallExpression) {
       if (node.isSubtypeOfNavigableCompat() && node.receiverType?.canonicalText == ACTIVITY)
```
ryanmoelter, line 36: nit: we probably want braces around this if statement

ryanmoelter (approved):
Good ol' ktlint

magellan-lint/src/main/java/com/wealthfront/magellan/AvoidUsingActivity.kt
narenkmanoharan: d

PR #149, opened 2020-10-14
Various fixes from the initial integration

narenkmanoharan:
This PR has a number of fixes that i found from the initial integration. I expect a lot more from when we do the actual integration this sprint.


magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyNavigator.kt
```diff
@@ -35,6 +36,9 @@ class LegacyNavigator internal constructor(
       if (navItem is Screen<*>) {
         navItem.setNavigator(this)
       }
+      if (navItem is MultiScreen<*>) {
+        navItem.screens.forEach { it.setNavigator(this) }
```
narenkmanoharan, line 40: This sets the navigator for the child screen in screen groups.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -46,10 +46,11 @@ internal class ActivityLifecycleAdapter(
...
-  navigable: Navigable,
-  @LayoutRes root: Int = R.layout.magellan_root
-) {
+fun ComponentActivity.setContentScreen(navigable: Navigable, @LayoutRes root: Int = R.layout.magellan_root) {
   setContentView(root)
   lifecycle.addObserver(ActivityLifecycleAdapter(navigable, this))
 }
+
+fun ComponentActivity.setExpedition(navigable: Navigable) {
+  lifecycle.addObserver(ActivityLifecycleAdapter(navigable, this))
```
narenkmanoharan, line 55: This method is necessary for us where we inflate the xml before the navigable is available through dependency injection

magellan-library/src/main/java/com/wealthfront/magellan/navigation/ActionBarConfigListener.java
```diff
@@ -3,9 +3,10 @@
 import com.wealthfront.magellan.view.ActionBarConfig;
 
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public interface ActionBarConfigListener {
 
-  void onNavigate(@NotNull ActionBarConfig actionBarConfig);
+  void onNavigate(@Nullable ActionBarConfig actionBarConfig);
```
narenkmanoharan, line 10: This method needs to be called for screen that actually done implement this interface as well. Steps don't automatically implement this interface.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableListener.java
```diff
@@ -14,12 +14,12 @@ default void onNavigableHidden(@NotNull NavigableCompat navigable) {}
   default void onNavigate() {}
 
   @Override
-  default void create(@NotNull Context context) {
+  default void show(@NotNull Context context) {
```
narenkmanoharan, line 17: this prevents leaks.

magellan-legacy/src/main/java/com/wealthfront/magellan/MultiScreen.java
```diff
@@ -0,0 +1,18 @@
+package com.wealthfront.magellan;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public interface MultiScreen<S extends Screen> {
```
ryanmoelter, line 7: I'm not against this at all, but curious: what prompted us to split this out into its own interface?

magellan-library/src/main/java/com/wealthfront/magellan/core/Step.kt
```diff
@@ -10,10 +10,11 @@ import androidx.annotation.VisibleForTesting
 import androidx.viewbinding.ViewBinding
 import com.wealthfront.magellan.lifecycle.LifecycleAwareComponent
 import com.wealthfront.magellan.lifecycle.lifecycleWithContext
+import com.wealthfront.magellan.view.ActionBarModifier
 
 abstract class Step<V : ViewBinding>(
   createBinding: (LayoutInflater) -> V
-) : Navigable, LifecycleAwareComponent() {
+) : Navigable, LifecycleAwareComponent(), ActionBarModifier {
```
ryanmoelter, line 17: I worry that we're piling too much into the `Step` and `Journey` classes, when we're trying to practice composition over inheritance. Should `Step`s really be `ActionBarModifier`s, or should they have one instead?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
ryanmoelter: Wait, when does this happen?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
```diff
@@ -44,6 +44,14 @@ abstract class LifecycleAwareComponent : LifecycleAware, LifecycleOwner {
...
   }
 
+  fun attachToLifecycle(lifecycleAware: LifecycleAware) {
+    attachToLifecycle(lifecycleAware, LifecycleState.Destroyed)
+  }
+
+  fun removeFromLifecycle(lifecycleAware: LifecycleAware) {
+    removeFromLifecycle(lifecycleAware, LifecycleState.Destroyed)
+  }
+
```
ryanmoelter, line 54: Why not just use default parameters for the existing attachToLifecycle methods? Wait, don't the existing attachToLifecycle methods have default parameters inherited from the interface?

If we're going to add methods like this, it should be to the `LifecycleAware` interface.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/CurrentNavigableProvider.kt
```diff
@@ -0,0 +1,16 @@
+package com.wealthfront.magellan.navigation
+
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CurrentNavigableProvider @Inject constructor() : NavigableListener {
+
+  lateinit var navigable: NavigableCompat
```
ryanmoelter, line 9: Should probably add a `private set` to this

magellan-library/src/main/java/com/wealthfront/magellan/view/ActionBarModifier.java
```diff
@@ -27,6 +27,13 @@ default boolean shouldAnimateActionBar() {
     return true;
   }
 
+  /**
+   * @return true if we need to pass the onUpdateMenu calls to the navigables children, false otherwise (false by default).
+   */
+  default boolean shouldShowChildNavigablesMenu() {
+    return false;
+  }
```
ryanmoelter, line 35: Remind me again how we propagate `onUpdateMenu` and populate the menu? Is it in `onUpdateMenu` above, i.e. the implementation is provided by the implementor of this interface?

This method is kind of weird with no implementation backing it. `shouldShowActionBar` above is acted on by another object, so that's exposing this state to another class. But overriding this method and expecting the inheritor of this class to respect it is a strange pattern.

Maybe we should add this to `LegacyScreen` or wherever the logic for passing the menu to its children actually lives.

ryanmoelter (changes requested):

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
narenkmanoharan: This is gonna be mostly for migration purposes and we will mark this as opt in...

magellan-legacy/src/main/java/com/wealthfront/magellan/MultiScreen.java
narenkmanoharan: Talked offline. This is fine.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
narenkmanoharan: Add a default param to LifecycleAware instead.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/CurrentNavigableProvider.kt
narenkmanoharan: yeah. can do.

magellan-library/src/main/java/com/wealthfront/magellan/view/ActionBarModifier.java
narenkmanoharan: Remove this.

magellan-library/src/main/java/com/wealthfront/magellan/core/Step.kt
narenkmanoharan: Remove the ActionBarModifier and make the onNavigate method have a nullable config. We will hide the actionbar when its null

ryanmoelter (approved):
LGTM 🦖

PR #150, opened 2020-10-20
Use Kotlin 1.4 explicit mode

narenkmanoharan:
This is a [pretty cool new feature](https://kotlinlang.org/docs/reference/whatsnew14.html#explicit-api-mode-for-library-authors) introduced recently in 1.4. This forces us to explicitly mention our return types and access modifiers so that we don't accidentally break them in the future.

We should also consider adding this to our library modules (Maybe UI-toolkit as well)


magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleState.kt
```diff
@@ -6,27 +6,27 @@ import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.BACKWARDS
...
+  internal fun getEarlierOfCurrentState(): LifecycleState = when (this) {
     is Destroyed -> Destroyed
     is Created, is Shown, is Resumed -> Created(context!!)
   }
 }
 
-enum class LifecycleStateDirection {
+internal enum class LifecycleStateDirection {
   FORWARD, BACKWARDS, NO_MOVEMENT;
 }
```
ryanmoelter, line 32: I forget where we use these, but would we need these public to reimplement `LinearNavigator`? If so, we should leave them public

ryanmoelter (approved):
This is cool! We'll have to use this more often.

Just a few questions before merging this, but overall, looks great!

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAware.java
```diff
@@ -0,0 +1,22 @@
+package com.wealthfront.magellan.lifecycle;
+
+import android.content.Context;
+
+public interface LifecycleAware {
```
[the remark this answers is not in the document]
narenkmanoharan: I'm trying to remove our reliance on the consumer of the libraries having to pass the [-Xjvm-default=compatibility] flag. But i might need to do more work there in a later PR.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleState.kt
narenkmanoharan: d.

PR #152, opened 2021-01-22
Add compatibility for history rewriting APIs and rename LegacyNavigator

narenkmanoharan:
This PR adds better backwards compatibility support for the HistoryRewriter APIs. This should make the migration simpler and support the old use cases better.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 GROUP=com.wealthfront
-VERSION_NAME=2.0.0-development
+VERSION_NAME=2.0.1-alpha
 
 POM_DESCRIPTION=The simplest navigation library for Android
 
--- a/magellan-legacy/src/main/java/com/wealthfront/magellan/HistoryRewriter.java
+++ b/magellan-legacy/src/main/java/com/wealthfront/magellan/HistoryRewriter.java
@@ -1,0 +1,17 @@
+package com.wealthfront.magellan;
+
+import android.app.Activity;
+
+import com.wealthfront.magellan.navigation.NavigableCompat;
+
+import java.util.Deque;
+
+public interface HistoryRewriter {
+
+  /**
+   * Used to rewrite history on the fly, using either {@link Navigator#rewriteHistory(Activity, HistoryRewriter)} or
+   * {@link Navigator#navigate(HistoryRewriter)} (and its variants).
+   */
+  void rewriteHistory(Deque<NavigableCompat> history);
+
+}
--- a/magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
+++ b/magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
@@ -1,5 +1,6 @@
 package com.wealthfront.magellan
 
+import android.app.Activity
 import android.view.Menu
 import com.wealthfront.magellan.Direction.BACKWARD
 import com.wealthfront.magellan.Direction.FORWARD
@@ -14,17 +15,17 @@ import com.wealthfront.magellan.transitions.DefaultTransition
 import com.wealthfront.magellan.transitions.MagellanTransition
 import com.wealthfront.magellan.transitions.NoAnimationTransition
 import com.wealthfront.magellan.transitions.ShowTransition
-import java.lang.IllegalStateException
-import java.util.Stack
+import rewriteHistoryWithNavigationEvents
+import java.util.Deque
 
-public class LegacyNavigator internal constructor(
+public class Navigator internal constructor(
   override val journey: Step<*>,
   container: () -> ScreenContainer
 ) : Navigator, LifecycleAwareComponent() {
 
   private val delegate by lifecycle(NavigationDelegate(journey, container))
 
-  override val backStack: Stack<NavigationEvent>
+  override val backStack: Deque<NavigationEvent>
     get() = delegate.backStack
 
   internal var menu: Menu? = null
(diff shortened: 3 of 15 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 GROUP=com.wealthfront
-VERSION_NAME=2.0.0-development
+VERSION_NAME=2.0.1-alpha
```
narenkmanoharan, line 2: sneaky fix: Update to the latest version

magellan-legacy/src/main/java/com/wealthfront/magellan/HistoryRewriter.java
```diff
@@ -0,0 +1,17 @@
+package com.wealthfront.magellan;
+
+import android.app.Activity;
+
+import com.wealthfront.magellan.navigation.NavigableCompat;
+
+import java.util.Deque;
+
+public interface HistoryRewriter {
```
narenkmanoharan, line 9: Bringing back some old classes for easier integration with our codebase.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -35,7 +36,7 @@ public class NavigationDelegate(
       updateMenu(menu)
     }
 
-  public val backStack: Stack<NavigationEvent> = Stack()
+  public val backStack: Deque<NavigationEvent> = ArrayDeque()
```
narenkmanoharan, line 39: Changed from Stack to ArrayDeque for backwards compatibility, there should be virtually no API difference..

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -12,8 +13,9 @@ public class NavigationTraverser(private val root: NavigableCompat) {
   private fun constructTree(navigable: NavigableCompat): NavigationNode {
     val navNode = NavigationNode(navigable, mutableListOf())
     if (navNode.hasBackstack) {
-      navNode.backStack.forEach { navEvent ->
-        val childNavNode = constructTree(navEvent.navigable)
+      while (navNode.backStack.isNotEmpty()) {
```
narenkmanoharan, line 16: Changes because of the switch from Stack to Queue. Go unit tests!

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
```diff
@@ -14,17 +15,17 @@ import com.wealthfront.magellan.transitions.DefaultTransition
 import com.wealthfront.magellan.transitions.MagellanTransition
 import com.wealthfront.magellan.transitions.NoAnimationTransition
 import com.wealthfront.magellan.transitions.ShowTransition
-import java.lang.IllegalStateException
-import java.util.Stack
+import rewriteHistoryWithNavigationEvents
+import java.util.Deque
 
-public class LegacyNavigator internal constructor(
+public class Navigator internal constructor(
```
ryanmoelter, line 21: What was the thinking behind renaming this again?

magellan-library/src/main/java/com/wealthfront/magellan/ScreenContainer.java
```diff
@@ -6,7 +6,7 @@
 import android.widget.FrameLayout;
 
 /**
- * The container to be used to display the screens using the {@link Navigator}. Must have the id
+ * The container to be used to display the screens using the {@link com.wealthfront.magellan.navigation.Navigator}. Must have the id
```
ryanmoelter, line 9: Nit: probably don't want the full classpath here

ryanmoelter: Oh wait, I thought it was kotlin docs, idk what the normal thing to do for javadocs is anymore haha

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
ryanmoelter: I forget how this is structured, but is there a way we can contain this change to the `magellan-legacy` artifact? I think `Stack` is a better model for what we're trying to do here and would like to use that moving forward if possible. I don't feel that strongly though.

ryanmoelter (approved):
LGTM with a couple questions :)

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
narenkmanoharan: Hmm. It would be weird to have different data structures for the same functionality between magellan-library and magellan-legacy. I was thinking about the pros and cons and there doesn't seem to be anything that swings me either way so I thought i can keep the migration and thinking simpler to keep the existing `ArrayDeque` that we were using in 1.0.

magellan-library/src/main/java/com/wealthfront/magellan/ScreenContainer.java
narenkmanoharan: The IDE sorta autocompleted with the full path so I just left it.

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
narenkmanoharan: So i wanted to keep the integration simple so this should help so that I don't have to change all the places in the app that imports the navigator for now. But I will probably change it back in a small update after the integration.

PR #154, opened 2021-02-10
Fix for double navigation on initialization and deep copy for logging backstack

narenkmanoharan:


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -54,8 +56,11 @@ public data class NavigationNode(
     get() = (value as? LifecycleOwner)?.children?.mapNotNull { it as? Navigator }?.isNotEmpty()
       ?: false
 
-  val backStack: Deque<NavigationEvent>
-    get() = (value as LifecycleOwner).children.mapNotNull { it as? Navigator }.first().backStack
+  public fun getBackstack(): Deque<NavigationEvent> {
+    val backStackDeepCopy = ArrayDeque<NavigationEvent>()
+    (value as LifecycleOwner).children.mapNotNull { it as? Navigator }.first().backStack.toList().toCollection(backStackDeepCopy)
```
narenkmanoharan, line 61: had to fix this since we moved to using ArrayDeque.

magellan-sample/src/main/java/com/wealthfront/magellan/sample/Expedition.kt
```diff
@@ -20,7 +20,7 @@ class Expedition : Journey<ExpeditionBinding>(
 
   override fun onCreate(context: Context) {
     appComponent.inject(this)
-    navigator.goTo(FirstJourney(::goToSecondJourney))
+    navigator.setRoot(FirstJourney(::goToSecondJourney))
```
narenkmanoharan, line 23: This was causing the onCreate to be triggered twice which we don't want and this fixes it :)

ryanmoelter (approved):
LGTM!

PR #155, opened 2021-03-05
Remove menu/toolbar related code out of magellan

narenkmanoharan:
We went back and forth on this and I think this is the right way to go after the initial integration. This complicates the library while trying to keep the subpar APIs, so I think its worth the investment to move the responsibility of handling the menu out of the library. This makes the library so much cleaner and focused (only on navigation). 

I added a sample of how we could handle menus in our app to the advanced sample. We can consider offering this in the support library when we eventually flesh this out.

- This clean up a lot of code from the core library and makes it focused only on navigation
- We don't have access to the activity anymore so we enforce that the context is only used between onShow and onHide (which is going to avoid a lot of edge cases)
- The setup of the library is much easier now. No need to hook up the menu and handle it in the main activity
- We can use a lifecycle/navigation aware object to handle this concern completely and be even more feature rich
- Managing the toolbar via dependency injection
- Can offer this in the `magellan-support` library
- More control about how everything is laid out inside the toolbar instead of relying on the native toolbar which is restrictive (i'm thinking better animations with blend etc.)
- No more god `menu.xml` that we need to maintain! :D
- Much more testable

Only con, I can think of is the migration to the new pattern, but I think this is a much better pattern and is worth the migration. Also, our batch file editor is going to come in handy.


magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
```diff
@@ -56,6 +47,7 @@ public class Navigator internal constructor(
     delegate.navigate(direction, backStackOperation)
   }
 
+  @JvmOverloads
```
narenkmanoharan, line 50: sneaky fix

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableCompat.kt
```diff
@@ -1,34 +1,6 @@
...
-
-  /**
-   * Set the title to be displayed on the navigable
-   */
-  @JvmDefault
-  public fun setTitle(title: CharSequence) {
-    activity?.title = title
-  }
-}
+public interface NavigableCompat : LifecycleAware, Displayable
```
ryanmoelter, line 6: Oh sweet! Can we remove this interface altogether in favor of the regular `Navigable`?

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/Expedition.kt
```diff
@@ -20,8 +21,13 @@ class Expedition @Inject constructor() : LegacyExpedition<ExpeditionBinding>(
   @set:Inject var navListener: LoggingNavigableListener by lateinitLifecycle()
 
   override fun onCreate(context: Context) {
+    attachToLifecycle(ToolbarHelper)
     app(context).injector().inject(this)
-    navigator.goTo(DogListStep(::goToDetailsScreen))
+  }
+
+  override fun onShow(context: Context, binding: ExpeditionBinding) {
+    ToolbarHelper.init(viewBinding!!.menu, navigator)
```
ryanmoelter, line 29: It makes sense to use a static class in the example, but do we want to add a comment explaining that you would usually use DI for this?

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/MainActivity.kt
```diff
@@ -29,51 +16,9 @@ class MainActivity : AppCompatActivity(), ActionBarConfigListener {
...
-  @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
-  private fun animateStatusBarColor(window: Window, startColor: Int, endColor: Int) {
-    if (startColor == endColor) {
-      return
-    }
-    val statusColorAnimator = ValueAnimator.ofObject(ArgbEvaluator(), startColor, endColor)
-    statusColorAnimator.duration = DEFAULT_ANIM_DURATION_MS
-    statusColorAnimator.addUpdateListener { window.statusBarColor = it.animatedValue as Int }
-    statusColorAnimator.start()
-  }
```
ryanmoelter, line 78: Yesssss move all of this out of the `MainActivity` and into a proper Step! This will make things much easier

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
```diff
@@ -0,0 +1,38 @@
+package com.wealthfront.magellan.sample.advanced.toolbar
+
+import android.content.Context
+import com.wealthfront.magellan.Navigator
+import com.wealthfront.magellan.navigation.NavigableCompat
+import com.wealthfront.magellan.navigation.NavigableListener
+
+object ToolbarHelper : NavigableListener {
```
ryanmoelter, line 8: I think I mentioned this in our conversation, but we should consider changing the API of this example to use a data class instead of multiple methods. That way, we configure all the options at once (and have a convenient map of all possible states for the toolbar)

ryanmoelter (commented):
Remove all the code! This looks great overall. Just a few minor comments

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/MainActivity.kt
narenkmanoharan: Yup!

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
narenkmanoharan: Agreed. This is the way to go. Will probably do this at a later point in time once we finalized the API and offer it through our support library.

ryanmoelter (approved):

PR #156, opened 2021-03-10
Add ability to disable animations, log in debug and change the default transition

narenkmanoharan:
These options allow for static customization of the library behavior similar to what we had in 1.0. Just decompiled it from the navigator to be global instead.


magellan-library/src/main/java/com/wealthfront/magellan/init/Magellan.kt
```diff
@@ -0,0 +1,33 @@
...
+    this.customDefaultTransition = defaultTransition
+  }
+}
+
+internal fun shouldRunAnimations(): Boolean = !disableAnimations
+
+internal fun shouldLogDebugInfo(): Boolean = logDebugInfo
+
+@RestrictTo(LIBRARY_GROUP_PREFIX)
+public fun getDefaultTransition(): MagellanTransition = customDefaultTransition
```
narenkmanoharan, line 33: This is used in magellan-legacy so needs to be public but it can only be used inside our libraries

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -127,7 +129,12 @@ public class NavigationDelegate(
   ) {
     currentNavigable!!.transitionStarted()
     to?.whenMeasured {
-      magellanTransition.animate(from, to, direction) {
+      val transition = if (shouldRunAnimations()) {
+        magellanTransition
+      } else {
+        NoAnimationTransition()
+      }
+      transition.animate(from, to, direction) {
```
ryanmoelter, line 116: Hmm, the only thing is that now all custom navigators will have to check this same field to know whether to run animations. Not necessarily a problem, but worth noting that we're making the Navigator contract more complex

ryanmoelter (approved):

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
narenkmanoharan: Thats fair. We can update the contract (interface) to denote that every consumer should probably support this.

PR #157, opened 2021-03-17
Minor fixes

narenkmanoharan:
This is just a collection of tiny tweaks that I found useful while doing the integration.

Improved backwards compatibility
- Added support for screen lifecycle listener
- Added support reset with root (useful for integration tests - only in legacy navigator)
- Added support for currentScreen & isCurrentScreen from the navigator (Through manual injection of the current navigable provider)
- Other tiny tweaks


magellan-legacy/src/main/java/com/wealthfront/magellan/HistoryRewriterExtensions.kt
```diff
@@ -24,8 +24,8 @@ internal fun HistoryRewriter.rewriteHistoryWithNavigationEvents(
     }
   }
   if (magellanTransition != null || navigationType != null) {
-    val lastNav = oldBackStack.last
-    oldBackStack.removeLast()
-    oldBackStack.add(NavigationEvent(lastNav.navigable, magellanTransition ?: navigationType!!.toTransition()))
+    val lastNav = oldBackStack.first
```
narenkmanoharan, line 27: sneaky fix

ryanmoelter: Wouldn't the last nav be at the top of the stack? Maybe we should call pop() here for clarity (is `first` the top or bottom of the stack?)

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
```diff
@@ -70,32 +77,8 @@ public class Navigator internal constructor(
...
-    }
-  }
-
-  @JvmOverloads
-  public fun hide(magellanTransition: MagellanTransition? = null) {
-    navigate(BACKWARD) { backStack ->
-      NavigationEvent(backStack.pop().navigable, magellanTransition ?: ShowTransition())
-    }
-  }
-
```
ryanmoelter, line 96: Don't we need these hide() functions for compatibility?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachine.kt
```diff
@@ -33,7 +33,7 @@ internal class LifecycleStateMachine {
       currentState = when (currentState.getTheDirectionIShouldGoToGetTo(newState)) {
         FORWARD -> next(subjects, currentState, newState.context!!)
         BACKWARDS -> previous(subjects, currentState, getContext(newState, oldState))
-        NO_MOVEMENT -> throw IllegalStateException("Attempting to transition from $currentState to itself")
+        NO_MOVEMENT -> newState
```
ryanmoelter, line 36: When does this happen? Wouldn't transitioning to the current state mean there's a problem somewhere?

magellan-legacy/src/main/java/com/wealthfront/magellan/ScreenLifecycleListener.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  @JvmDefault
+  override fun show(context: Context) {
+    addNavigableListener(this)
+  }
+
+  @JvmDefault
+  override fun hide(context: Context) {
+    removeNavigableListener(this)
+  }
```
ryanmoelter, line 33: Doesn't this get attached in `create`/`destroy` of `NavigableListener`? Why do we need to do this again?

narenkmanoharan: Good point. Removed

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachine.kt
narenkmanoharan: Yeah. Im not sure about this. Im gonna add some logging maybe. But removed it since it was not harmful.

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
narenkmanoharan: I moved all our hides to just use goBack instead. This way you have just one way of navigating and navigating back is automatically handled for you.

magellan-legacy/src/main/java/com/wealthfront/magellan/HistoryRewriterExtensions.kt
narenkmanoharan: Done. Changed to pop.

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
ryanmoelter: Ooooh sweet :)

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleStateMachine.kt
ryanmoelter: Could we keep it until it causes problems? I'd rather fail fast than have this in an unexpected state, personally

narenkmanoharan: Oh, i think i figured it out. We previously were comparing the instances of the data classes in the while loop. Switched it over to compare the order instead. I think that should resolve the issue. Keeping the throw statement.

ryanmoelter (approved):
Just a few minor comments, but overall LGTM!

magellan-library/src/main/java/com/wealthfront/magellan/core/Journey.kt
```diff
@@ -12,6 +12,4 @@ public abstract class Journey<V : ViewBinding>(
 ) : Step<V>(createBinding) {
 
   protected var navigator: LinearNavigator by lifecycle(LinearNavigator { viewBinding!!.container() })
-
-  override fun toString(): String = this.javaClass.simpleName
```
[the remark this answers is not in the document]
narenkmanoharan: This was just so that we don't have a memory address printed in the logs during navigation but i've actually found them pretty useful so removing it so that we can have the class name + memory address

gradle/gradle-mvn-push.gradle
```diff
@@ -142,7 +142,7 @@ afterEvaluate { project ->
   }
 
   signing {
-    publishing.publications.all { publication ->
+    publishing.publications.all {publication ->
```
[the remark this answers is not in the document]
narenkmanoharan: ugh. Android studio. Done

PR #159, opened 2021-04-02
Add mockable to make classes testable

narenkmanoharan:


magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
```diff
@@ -99,14 +109,28 @@ public class Navigator internal constructor(
 
   public fun isCurrentScreen(other: NavigableCompat): Boolean = currentNavigableProvider!!.isCurrentNavigable(other)
 
-  public fun currentScreen(): NavigableCompat = currentNavigableProvider!!.navigable!!
+  public fun currentScreen(): NavigableCompat? = currentNavigableProvider!!.navigable
```
narenkmanoharan, line 112: Made `currentScreen` nullable to reflect the magellan 2.0 change and also added a couple of utility method for making the migration easier

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
```diff
@@ -54,26 +55,35 @@ public class Navigator internal constructor(
     delegate.navigate(direction, backStackOperation)
   }
 
-  @JvmOverloads
-  public fun replace(navigable: NavigableCompat, magellanTransition: MagellanTransition? = null) {
```
ryanmoelter, line 58: Do default parameters cause problems for mocking? I think I remember something about that

ryanmoelter (approved):
I know that most of this is for backwards compatibility, but I still don't like the idea of opening all of this up just for testing. In the future, we should probably provide fakes for testing

magellan-legacy/src/main/java/com/wealthfront/magellan/Navigator.kt
narenkmanoharan: yeah. that was it. So I decided to break them apart into separate cascading methods instead of using `JvmOverloads` which doesn't seem to be working well with mocking

PR #165, opened 2021-04-20
Fix navigation delegate

ryanmoelter:
Whoops, this is getting pretty big. Here's what's here:

- Refactor NavigationDelegate to be based on LifecycleLimiter
- Fix some issues with LifecycleLimiter
  - Properly transition states
  - Only pass backPressed() events to Shown or Resumed children
- Add assertions for transitioning lifecycles from unexpected states and attaching or detaching multiple times
- Fix a bug in the StatePrinter

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/Direction.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/Direction.kt
@@ -5,7 +5,7 @@ package com.wealthfront.magellan
  */
 public enum class Direction(private val sign: Int) {
 
-  FORWARD(1), NO_MOVEMENT(0), BACKWARD(-1);
+  FORWARD(1), BACKWARD(-1);
 
   internal fun sign(): Int {
     return sign
--- a/magellan-library/src/main/java/com/wealthfront/magellan/debug/StatePrinter.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/debug/StatePrinter.kt
@@ -42,7 +42,7 @@ private fun LifecycleAware.getLifecycleStateSnapshotRecursive(
           "$VERTICAL_LINE$INDENT_SPACE"
         }
         lifecycleAware.getLifecycleStateSnapshotRecursive(
-          indent + childIndent,
+          childIndent,
           index == children.lastIndex,
           currentState
         )
--- a/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
@@ -15,7 +15,8 @@ public abstract class LifecycleAwareComponent : LifecycleAware, LifecycleOwner {
   final override fun create(context: Context) {
     if (currentState !is LifecycleState.Destroyed) {
       throw IllegalStateException(
-        "Cannot create() from a state that is not Destroyed: ${currentState::class.java.simpleName}"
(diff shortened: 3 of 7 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -97,12 +103,30 @@ public class NavigationDelegate(
     containerView?.setInterceptTouchEvents(true)
     navigationPropagator.beforeNavigation()
     val from = navigateFrom(currentNavigable, direction)
+    val oldBackStack = backStack.map { it.navigable }
     val transition = backStackOperation.invoke(backStack).magellanTransition
+    diffBackstackAndUpdateMaxStates(oldBackStack = oldBackStack, newBackStack = backStack.map { it.navigable })
```
narenkmanoharan, line 108: nit: maybe create a variable `newBackStack `

magellan-library/src/main/java/com/wealthfront/magellan/debug/StatePrinter.kt
```diff
@@ -42,7 +42,7 @@ private fun LifecycleAware.getLifecycleStateSnapshotRecursive(
           "$VERTICAL_LINE$INDENT_SPACE"
         }
         lifecycleAware.getLifecycleStateSnapshotRecursive(
-          indent + childIndent,
+          childIndent,
```
ryanmoelter, line 45: Sneaky fix: double indenting

magellan-library/src/main/java/com/wealthfront/magellan/Direction.kt
```diff
@@ -5,7 +5,7 @@ package com.wealthfront.magellan
  */
 public enum class Direction(private val sign: Int) {
 
-  FORWARD(1), NO_MOVEMENT(0), BACKWARD(-1);
+  FORWARD(1), BACKWARD(-1);
```
ryanmoelter, line 8: Don't need it anymore!

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
```diff
@@ -15,17 +19,27 @@ internal class LifecycleRegistry : LifecycleAware {
       val oldState = field
       field = newState
       listenersToMaxStates.forEach { (lifecycleAware, maxState) ->
-        if (newState.isWithinLimit(maxState)) {
-          lifecycleStateMachine.transition(lifecycleAware, oldState, newState)
```
ryanmoelter, line 19: This caused some double lifecycle transitions

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
```diff
@@ -13,31 +13,67 @@ public abstract class LifecycleAwareComponent : LifecycleAware, LifecycleOwner {
     get() = lifecycleRegistry.currentState
 
   final override fun create(context: Context) {
+    if (currentState !is LifecycleState.Destroyed) {
```
narenkmanoharan, line 16: thanks for adding these preconditions!

narenkmanoharan:
This looks great!!!

The only concern that I have is that the library is pretty complicated in terms on the internal implementation. It's still pretty simple to use but internally there's a lot of things going on. Might be good to have a conversation about that.

narenkmanoharan (approved):

PR #177, opened 2021-05-02
Spike compose support

ryanmoelter:
This is the gist of my idea for supporting compose. Still a ways to go. :)

Done:

- Create ComposeStep
- Create some compose <-> View interop support

Not done:

- Create compose-friendly `Navigator` and `Journey`
- See if we can keep the `minSdkLevel` low for non-compose packages
- Add tests


magellan-sample/src/main/java/com/wealthfront/magellan/sample/IntroStep.kt
```diff
@@ -1,16 +1,29 @@
...
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Devices
+import androidx.compose.ui.tooling.preview.Preview
+import com.example.magellan.compose.ComposeStep
 
 internal class IntroStep(
   private val goToLearnMore: () -> Unit
-) : Step<IntroBinding>(IntroBinding::inflate) {
+) : ComposeStep() {
```
ryanmoelter, line 12: The first `ComposeStep`!

magellan-sample/src/main/java/com/wealthfront/magellan/sample/IntroStep.kt
```diff
@@ -1,16 +1,29 @@
...
+  @Composable
+  override fun Compose() = IntroStepContent(goToLearnMore)
+}
+
+@Composable
+internal fun IntroStepContent(goToLearnMore: () -> Unit) {
+  Button(onClick = goToLearnMore) {
+    Text("Go to learn more")
   }
 }
```
ryanmoelter, line 23: On second thought, this probably belongs inside the `Step`. That way, it can access the data inside the `Step`, which we can fake for the previews.

```kotlin
class MyStep : ComposeStep() {
  val myDataFlow = StateFlow<MyData>()

  @Composable
  override fun Compose() {
    val myData = myDataFlow.collectAsState()
    // Use myData here
  }
}
```

magellan-compose/src/main/java/com/example/magellan/compose/SimpleComposeStep.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import androidx.compose.runtime.Composable
+
+public class SimpleComposeStep(public val Content: @Composable SimpleComposeStep.() -> Unit) : ComposeStep() {
+
+  @Composable
+  override fun Compose() {
+    this.Content()
+  }
+}
```
ryanmoelter, line 11: This could be useful for static content, but that doesn't happen often.

buildSrc/build.gradle
```diff
@@ -9,18 +9,22 @@ repositories {
 }
 
 dependencies {
-  implementation "com.android.tools.build:gradle:4.1.1"
+  implementation 'com.android.tools.build:gradle:7.0.0-beta05'
+  // See https://issuetracker.google.com/issues/176079157#comment11
+  implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.10"
```
ryanmoelter, line 14: Figuring this out took literally TWO AND A HALF MONTHS. Everything else worked from the start.

narenkmanoharan: That's pretty annoying.

buildSrc/build.gradle
```diff
@@ -9,18 +9,22 @@ repositories {
 }
 
 dependencies {
-  implementation "com.android.tools.build:gradle:4.1.1"
+  implementation 'com.android.tools.build:gradle:7.0.0-beta05'
```
narenkmanoharan, line 12: nit: Since AS arctic fox has been released, we can bump this to the stable version. Same with compose 🥳

magellan-compose/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+    jvmTarget = "1.8"
+    useIR = true
+  }
+
+  composeOptions {
+    kotlinCompilerExtensionVersion Versions.composeVersion
+  }
+
+  compileOptions {
+    setSourceCompatibility(JavaVersion.VERSION_1_8)
```
narenkmanoharan, line 34: Might need to bump this to Java 11 for compatibility with arctic fox.

magellan-sample/src/main/java/com/wealthfront/magellan/sample/IntroStep.kt
```diff
@@ -1,16 +1,29 @@
...
+
+@Preview(device = Devices.PIXEL_3, name = "Intro step")
+@Composable
```
narenkmanoharan, line 26: We should probably encourage the pattern on annotating the compose function with the preview annotation so that you can look at how the entire step view will look like and if we want to look at other smaller functions we can do that too. Creating a separate method just for the preview might create misdirection when reading the code.

magellan-compose/src/main/java/com/example/magellan/compose/ComposeStep.kt
```diff
@@ -0,0 +1,21 @@
+package com.example.magellan.compose
```
narenkmanoharan, line 1: nit: the naming of the classes is a bit confusing. But I can't think of better names. Something we might want to revisit later.

magellan-compose/src/main/java/com/example/magellan/compose/ComposeInterop.kt
```diff
@@ -0,0 +1,49 @@
...
+      throw IllegalStateException(
+        "View does not exist on ${displayable::class.java.simpleName}. " +
+          "Is it attached to the lifecycle?"
+      )
+    }
+    displayable.view!!
+  })
+}
+
+public class ComposeStepWrapper(
```
narenkmanoharan, line 27: nit: the naming of the classes is a bit confusing. But I can't think of better names. Something we might want to revisit later.

narenkmanoharan (approved):

PR #184, opened 2021-07-15
Fix overlapping activities bug

ryanmoelter:
Should fix the bug we're seeing around one activity being started before the other one's torn down.

Unfortunately, gradle isn't working on my m1 mac so I can't run it right now. I added tests in another PR so we can merge this and get those working when I have my laptop again.

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
@@ -1,13 +1,19 @@
 package com.wealthfront.magellan.lifecycle
 
 import android.app.Activity
+import android.content.Context
 import androidx.activity.ComponentActivity
 import androidx.annotation.LayoutRes
 import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.Lifecycle as ActivityLifecycle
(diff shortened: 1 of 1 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/ActivityLifecycleAdapter.kt
```diff
@@ -1,13 +1,19 @@
...
+import androidx.lifecycle.Lifecycle as ActivityLifecycle
+import androidx.lifecycle.LifecycleOwner as ActivityLifecycleOwner
 import com.wealthfront.magellan.R
 import com.wealthfront.magellan.ScreenContainer
 import com.wealthfront.magellan.core.Navigable
-import androidx.lifecycle.LifecycleOwner as ActivityLifecycleOwner
+import com.wealthfront.magellan.lifecycle.LifecycleState.Created
+import com.wealthfront.magellan.lifecycle.LifecycleState.Destroyed
+
+private var adapterMap = emptyMap<Navigable, Pair<ActivityLifecycleAdapter, ActivityLifecycle>>()
```
narenkmanoharan, line 16: nit: rename to ActivityLifecycleMap or something like that. Feel free to do in a separate PR as well or when you get to the java doc.

narenkmanoharan (approved):

PR #189, opened 2021-07-27
Null-out view reference in onStop

cmathew:
this should fix memory leaks we see when the Activity gets torn down by the system


ryanmoelter (approved):
Thanks for catching this!

narenkmanoharan (approved):

PR #196, opened 2021-09-09
Rename onShow to onStart in M2.0

cmathew:
IMO this is more accurate/intuitive with our Journey/Step verbiage (e.g. start a Journey vs show a Journey). Keep onShow/onHide in relevant places for 1.0 compat

### DIFF
--- a/magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyJourney.kt
+++ b/magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyJourney.kt
@@ -6,7 +6,7 @@ import com.wealthfront.magellan.core.Step
 import com.wealthfront.magellan.lifecycle.attachFieldToLifecycle
 import com.wealthfront.magellan.navigation.CurrentNavigableProvider
 
-public abstract class LegacyExpedition<V : ViewBinding>(
+public abstract class LegacyJourney<V : ViewBinding>(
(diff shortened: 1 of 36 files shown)


magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyJourney.kt
```diff
@@ -6,7 +6,7 @@ import com.wealthfront.magellan.core.Step
 import com.wealthfront.magellan.lifecycle.attachFieldToLifecycle
 import com.wealthfront.magellan.navigation.CurrentNavigableProvider
 
-public abstract class LegacyExpedition<V : ViewBinding>(
+public abstract class LegacyJourney<V : ViewBinding>(
```
ryanmoelter, line 9: I like this rename 👍

ryanmoelter (approved):
I'm pretty torn on this TBH. I like that we're borrowing names from the android framework, especially since `show`/`hide`/`start`/`stop` are called at the same time as their `Activity` counterparts. It is a break from 1.0 though, and that means context switching in legacy codebases, but technically we already have to do that between `onShow`/`onSubscirbe`.

So overall, I'm for it. Might be worth checking with the rest of the team though.

narenkmanoharan (approved):
This actually highlights the difference between onShow in 1.0 and 2.0 pretty well.

PR #197, opened 2021-09-09
Fix Advanced Sample compilation

cmathew:


magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
```diff
@@ -25,16 +23,6 @@ object ToolbarHelper : LifecycleAwareComponent() {
...
   }
 
-  override fun onStart(context: Context) {
-    shownScope.launch {
-      NavigationPropagator.events
-        .filterIsInstance<NavigationLifecycleEvent.NavigatedFrom>()
-        .collect { event ->
-          toolbarView?.reset()
-        }
-    }
```
ryanmoelter, line 35: We should replace this with a call to `NavigationPropagator.addListener(...)` (or whatever the API is) and removing it in `onStop`

ryanmoelter (approved):

PR #198, opened 2021-09-09
Bump test libs to support Android 12

cmathew:
So that test activities have Android 12-friendly export settings. Namely avoids errors like:
```
/Users/chrismathew/dev/Ferdinand/app/src/main/AndroidManifest.xml Error:
	android:exported needs to be explicitly specified for <activity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
```

### DIFF
--- a/buildSrc/src/main/kotlin/Versions.kt
+++ b/buildSrc/src/main/kotlin/Versions.kt
@@ -1,6 +1,6 @@
 object Versions {
   const val compileSdkVersion = 30
-  const val minSdkVersion = 15
+  const val minSdkVersion = 18
   const val targetSdkVersion = 30
 
   const val kotlinVersion = "1.4.30"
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -26,15 +26,14 @@ object Versions {
...
+  const val junitVersion = "4.13.2"
+  const val junitTestExtVersion = "1.1.3"
   const val truthVersion = "1.0"
   const val mockitoVersion = "2.23.4"
-  const val testRunnerVersion = "1.0.2"
+  const val testRunnerVersion = "1.4.0"
   const val uiAutomatorVersion = "2.2.0"
-  const val extVersion = "1.1.1"
-  const val espressoVersion = "3.0.2"
+  const val espressoVersion = "3.4.0"
```
cmathew, line 36: this library pulls in BootstrapActivity, EmptyActivity, and EmptyFloatingActivity. It's important to use newer versions of the lib so that we get this commit: https://github.com/android/android-test/commit/dd6df4aa22b77290fbe45834619217fd4852ba05

ryanmoelter (approved):

narenkmanoharan (approved):

magellan-sample/src/androidTest/java/com/wealthfront/magellan/uitest/DialogTest.kt
```diff
@@ -22,12 +24,19 @@ class DialogTest {
...
     onView(isRoot()).perform(orientationLandscape())
 
-    assertShown { text("Hello") }
-    assertShown { text("Are you sure about this?") }
+    assertDialogContentsShown()
   }
 }
+
+private fun assertDialogContentsShown() {
+  onView(withText("Hello"))
```
cmathew, line 37: I factored this out because thought I was going to need to target a dialog root for these matchers, but turns out I didn't have to...leaving it this 🤷

PR #199, opened 2021-09-10
Fix most unit tests

cmathew:
This fixes all but com.wealthfront.magellan.navigation.NavigationTraverserTest#globalBackStackWithSiblingJourney, which I'll look at in a separate PR.

- Factor out useful test components (DummyStep, DummyActivity, DummyScreen) into helper module with its own manifest
- Fix robolectric usage in unit tests to get them passing again


magellan-library-test/src/main/java/AttachingActivity.kt
```diff
@@ -0,0 +1,17 @@
+package com.wealthfront.magellan
+
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import com.wealthfront.magellan.core.Step
+import com.wealthfront.magellan.lifecycle.setContentScreen
+
+public class AttachingActivity : ComponentActivity() {
+  public companion object {
+    public var step: Step<*>? = null
```
cmathew, line 10: @ryanmoelter would you like me to change this to Navigable?

settings.gradle
```diff
@@ -7,3 +7,4 @@ include ':magellan-sample'
 include ':magellan-sample-advanced'
 include ':magellan-support'
 include ':magellan-test'
+include ':magellan-library-test'
```
cmathew, line 10: this new module contains objects that are helpful when writing unit tests for the Magellan lib itself. It's not exposed to people who pull in the library. Thoughts on naming? I also considered nesting like we do in wf-android: `:verification:library-test`

settings.gradle
```diff
@@ -7,3 +7,4 @@ include ':magellan-sample'
 include ':magellan-sample-advanced'
 include ':magellan-support'
 include ':magellan-test'
+include ':library-test'
```
cmathew, line 10: this new module contains objects that are helpful when writing unit tests for the Magellan lib itself. It's not exposed to people who pull in the library. Thoughts on naming? I also considered nesting like we do in wf-android e.g. `:verification:library-test`

PR #200, opened 2021-09-10
Merge LifecycleLimiter into LifecycleRegistry, expose LifecycleRegistry

cmathew:
this simplifies the code and makes it possible for others to build their own version of LifecycleAwareComponent if they so choose.


magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
```diff
@@ -6,11 +6,11 @@ import com.wealthfront.magellan.lifecycle.LifecycleLimit.DESTROYED
 import com.wealthfront.magellan.lifecycle.LifecycleLimit.NO_LIMIT
 import com.wealthfront.magellan.lifecycle.LifecycleLimit.SHOWN
 
-internal class LifecycleRegistry : LifecycleAware {
+public class LifecycleRegistry : LifecycleAware {
```
cmathew, line 9: had to make this public to match LifecycleLimiter. IDK we are worried about exposing this

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -31,8 +29,6 @@ public class NavigationDelegate(
   private val navigationPropagator = NavigationPropagator
   public val backStack: Deque<NavigationEvent> = ArrayDeque()
 
-  private val lifecycleLimiter by attachFieldToLifecycle(LifecycleLimiter())
```
cmathew, line 34: this is what spurred me to look into this change. Since NavigationDelegate extends LifecycleAwareComponent, it already has a LifecycleRegistry. Creating a lifecycleLimiter on top of that resulted in two `LifecycleRegistry`s for each NavigationDelegate

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
ryanmoelter: I think if we're going to expose this already (which seems reasonable to me), we should just get rid of `LifecycleLimiter` altogether and just use `LifecycleRegistry` directly. Any methods that only exist in `LifecycleLimiter` now can either move into `LifecycleLimiter` directly or be extension functions on it following [extension oriented design](https://elizarov.medium.com/extension-oriented-design-13f4f27deaee)

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleLimiter.kt
```diff
@@ -4,9 +4,9 @@ import android.content.Context
 import com.wealthfront.magellan.OpenForMocking
 
 @OpenForMocking
-public class LifecycleLimiter : LifecycleOwner, LifecycleAware {
-
-  private val lifecycleRegistry = LifecycleRegistry()
+public class LifecycleLimiter(
+  private val lifecycleRegistry: LifecycleRegistry
+) : LifecycleOwner, LifecycleAware {
```
ryanmoelter, line 9: `LifecycleRegistry`s are pretty cheap right? I'm not sure I like the pattern of passing `LifecycleRegistry`s around, it kind of breaks the encapsulation that we otherwise have.

One subtle bug that could show up as a result is the order in which children receive events. If another object (like LifecycleLimiter here) can muck around in registry, it can add things in an unpredictable order. Since we dispatch lifecycle events to children in the order they're attached to the lifecycle, this can cause unpredictable behavior.

Since `LifecycleRegistry`s are pretty cheap, we should consider encouraging people to use those rather than break encapsulation IMO.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleLimiter.kt
```diff
@@ -44,11 +44,15 @@ public class LifecycleLimiter : LifecycleOwner, LifecycleAware {
     attachToLifecycleWithMaxState(lifecycleAware, LifecycleLimit.NO_LIMIT, detachedState)
   }
 
-  public fun attachToLifecycleWithMaxState(lifecycleAware: LifecycleAware, maxState: LifecycleLimit, detachedState: LifecycleState = LifecycleState.Destroyed) {
+  internal fun attachToLifecycleWithMaxState(
```
ryanmoelter, line 47: Curious of the thinking behind making this `internal`? Do we expect people consuming the library not to use this method? (We use it ourselves in the `DashboardTabsScreen` right now)

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleAwareComponent.kt
```diff
@@ -5,6 +5,7 @@ import android.content.Context
 public abstract class LifecycleAwareComponent : LifecycleAware, LifecycleOwner {
 
   private val lifecycleRegistry = LifecycleRegistry()
+  protected val lifecycleLimiter: LifecycleLimiter = LifecycleLimiter(lifecycleRegistry)
```
ryanmoelter, line 8: Another idea: if we only want to expose some of the functionality of the `LifecycleRegistry` and not the whole class, what if we made `LifecycleLimiter` an interface and exposed it here under that interface?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
```diff
@@ -29,7 +29,7 @@ internal class LifecycleRegistry : LifecycleAware {
       }
     }
 
-  fun attachToLifecycle(
+  internal fun attachToLifecycle(
```
ryanmoelter, line 32: I'm curious what the value is of having this class `public` when pretty much all of its methods are `internal`. Care to share your thoughts?

ryanmoelter (changes requested):
We should figure out a few of these visibility and encapsulation questions before merging this, IMO

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
cmathew: Hah you're right, this is dumb

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleRegistry.kt
cmathew: Good idea. I ended up merging them under `LifecycleRegistry`

PR #202, opened 2021-09-14
Prevent multiple instances of the same Navigable on the backstack

ryanmoelter:
Having the same navigable in multiple places in the backstack caused an issue in our main app earlier. Even before this PR, we enforce that you can attach a `LifecycleAware` to a given `LifecycleOwner` at most once, but the way we detect changes in the backstack (so we can properly attach/detach navigables) means that we didn't properly enforce that for `NavigationDelegate`. Now, we will crash if you try to add a single instance to the backstack multiple times.

(This has been sitting on my laptop for a while, whoops. 😅)

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
@@ -120,6 +120,13 @@ public class NavigationDelegate(
   ) {
     val oldNavigables = oldBackStack.toSet()
     val newNavigables = newBackStack.toSet()
+    if (newNavigables.size != newBackStack.size) {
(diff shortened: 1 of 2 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -120,6 +120,13 @@ public class NavigationDelegate(
   ) {
     val oldNavigables = oldBackStack.toSet()
     val newNavigables = newBackStack.toSet()
+    if (newNavigables.size != newBackStack.size) {
```
cmathew, line 123: nice solution!

cmathew (approved):

PR #204, opened 2021-09-15
Add better testing support

ryanmoelter:
- Make `Journey.navigator` `@VisibleForTesting` so we can access it in tests for verification
- Make `transition` methods public, move from `LifecycleStateMachine` to extension functions on `LifecycleAware`
  - This makes it easier to get `LifecycleAware` objects into the right state when testing them
  - Also updated some sample tests to use this method instead of calling lifecycle events directly
- Detach `LifecycleAware` objects attached using `by attachFieldToLifecycle` when field is set (usually in tests)
  - This means that the original implementations don't get lifecycle events when we use fakes/mocks in tests


.idea/codeStyles/Project.xml
```diff
@@ -54,15 +54,6 @@
...
       </option>
-      <option name="PACKAGES_IMPORT_LAYOUT">
-        <value>
-          <package name="" alias="false" withSubpackages="true" />
-          <package name="java" alias="false" withSubpackages="true" />
-          <package name="javax" alias="false" withSubpackages="true" />
-          <package name="kotlin" alias="false" withSubpackages="true" />
-          <package name="" alias="true" withSubpackages="true" />
-        </value>
-      </option>
```
ryanmoelter, line 65: Sigh... Should we just keep this change, if AS wants to do it so bad?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleExtensions.kt
```diff
@@ -2,40 +2,63 @@ package com.wealthfront.magellan.lifecycle
...
+    hasOverrideValue = true
+    if (overrideValue != value) {
+      if (overrideValue is LifecycleAware) {
+        parent.removeFromLifecycle(overrideValue as LifecycleAware)
+      }
+      overrideValue = value
+      if (value is LifecycleAware) {
+        parent.attachToLifecycle(value)
+      }
+    }
```
ryanmoelter, line 61: This is more complicated because there are a few things happening here:
- Detach `lifecycleAware` if it hasn't already been overridden. We need to avoid detaching it twice
- We use a separate boolean `hasOverrideValue` to keep track, since `PropertyType` can be nullable and therefore `null` is a valid overriding value
- We detach any existing overriding value from the lifecycle for the same reason we detach `lifecycleAware`
  - Not all overriding values are `LifecycleAware` though, so we do this conditionally
- We finally attach the new overriding value, if it's `LifecycleAware`

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleTransitionExtensions.kt
```diff
@@ -0,0 +1,77 @@
...
+import android.content.Context
+import com.wealthfront.magellan.lifecycle.LifecycleState.Created
+import com.wealthfront.magellan.lifecycle.LifecycleState.Destroyed
+import com.wealthfront.magellan.lifecycle.LifecycleState.Resumed
+import com.wealthfront.magellan.lifecycle.LifecycleState.Shown
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.BACKWARDS
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.FORWARD
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.NO_MOVEMENT
+
+public fun <T> T.transitionToState(newState: LifecycleState) where T : LifecycleAware, T : LifecycleOwner {
```
ryanmoelter, line 12: See, generics can be fun 😉

I named it differently from `transition` below because I don't want people to get them confused, but I'm open to suggestions

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleTransitionExtensions.kt
```diff
@@ -0,0 +1,77 @@
...
+public fun Iterable<LifecycleAware>.transition(oldState: LifecycleState, newState: LifecycleState) {
+  var currentState = oldState
+  while (currentState.order != newState.order) {
+    currentState = when (currentState.getDirectionForMovement(newState)) {
+      FORWARD -> next(this, currentState, newState.context!!)
+      BACKWARDS -> previous(this, currentState, oldState.context!!)
+      NO_MOVEMENT -> throw IllegalStateException("Attempting to transition from $currentState to $newState")
+    }
+  }
+}
```
ryanmoelter, line 29: I took these out of LifecycleStateMachine because that class wasn't actually giving us any value, and this way is much more intuitive IMO

magellan-library/src/main/java/com/wealthfront/magellan/core/Journey.kt
```diff
@@ -11,5 +13,5 @@ public abstract class Journey<V : ViewBinding>(
   getContainer: V.() -> ScreenContainer
 ) : Step<V>(inflateBinding) {
 
-  protected var navigator: LinearNavigator by attachFieldToLifecycle(LinearNavigator { viewBinding!!.getContainer() })
+  @VisibleForTesting(otherwise = PROTECTED) public var navigator: LinearNavigator by attachFieldToLifecycle(LinearNavigator { viewBinding!!.getContainer() })
```
cmathew, line 16: nice! 

Random but in case you didn't know, there also something called RestrictTo: https://developer.android.com/studio/write/annotations#restrict. I learned about that somewhat recently so thought I'd mention it

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleTransitionExtensions.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.wealthfront.magellan.lifecycle.LifecycleState.Shown
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.BACKWARDS
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.FORWARD
+import com.wealthfront.magellan.lifecycle.LifecycleStateDirection.NO_MOVEMENT
+
+public fun <T> T.transitionToState(newState: LifecycleState) where T : LifecycleAware, T : LifecycleOwner {
+  transition(this.currentState, newState)
+}
+
+public fun LifecycleAware.transition(oldState: LifecycleState, newState: LifecycleState) {
```
cmathew, line 16: I must be thinking about this backwards. Aren't `LifecycleOwner`s the ones with state, shouldn't they be transitioning rather than `LifecycleAware`s? 

Is this simulating how "parents" drive their "child"s lifecycle?

cmathew (approved):

magellan-library/src/main/java/com/wealthfront/magellan/core/Journey.kt
ryanmoelter: Ooh good to know. In this case, we want it to be `protected | tests`, so I think `@VisibleForTesting` is what we want. But good to have another tool in the toolbox!

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleExtensions.kt
```diff
@@ -2,40 +2,63 @@ package com.wealthfront.magellan.lifecycle
 
 import kotlin.reflect.KProperty
 
-public fun <ChildType : LifecycleAware, PropertyType> LifecycleOwner.attachFieldToLifecycle(lifecycleAware: ChildType, getter: (ChildType) -> PropertyType): AttachFieldToLifecycleDelegate<ChildType, PropertyType> {
+public fun <ChildType : LifecycleAware, PropertyType> LifecycleOwner.attachFieldToLifecycle(
```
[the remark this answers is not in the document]
ryanmoelter: Well, technically, we are attaching the backing field to the lifecycle, and exposing a property computed from that field. I'm not opposed to coming up with new names, but I'm okay leaving the names how they are personally.

Though what do you think of `attachChildToLifecycle`?

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleExtensions.kt
```diff
@@ -2,40 +2,63 @@ package com.wealthfront.magellan.lifecycle
...
 
   public operator fun setValue(thisRef: Any?, property: KProperty<*>, value: PropertyType) {
-    overrideValue = value
+    if (!hasOverrideValue) {
+      parent.removeFromLifecycle(lifecycleAware)
+    }
+    hasOverrideValue = true
+    if (overrideValue != value) {
+      if (overrideValue is LifecycleAware) {
+        parent.removeFromLifecycle(overrideValue as LifecycleAware)
```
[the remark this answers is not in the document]
ryanmoelter: We remove `lifecycleAware`, i.e. the thing we attached from the constructor, on line 50. This is removing `overrideValue` (if it's not null), i.e. the value we may have set over the original value.

Tbh I'm not happy with `overrideValue` as a name, it clashes with the `override` keyword. But I also can't think of a better name

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleTransitionExtensions.kt
ryanmoelter: This is actually the function parents (well, technically `LifecycleRegistry`) call to move their children through their lifecycles, not even just simulating it.

I'm open to renaming things, but currently `LifecycleAware` is the interface for receiving events, therefore that that's the one we "drive" here by providing lifecycle events.

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/LifecycleExtensions.kt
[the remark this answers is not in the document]
cmathew: Ehh the names are fine, I was just confused heh

PR #209, opened 2021-09-21
Fix Advanced Sample compilation

cmathew:


magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
```diff
@@ -1,44 +1,39 @@
...
-        }
-    }
+  override fun onHide(context: Context) {
+    NavigationPropagator.removeNavigableListener(this)
   }
 
-  override fun onHide(context: Context) {
-    toolbarView = null
+  fun onDestroy() {
+    this.toolbarView = null
```
ryanmoelter, line 32: We should probably null this out in `onHide` rather than `onDestroy`

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
```diff
@@ -1,44 +1,39 @@
...
-  private val shownScope by attachFieldToLifecycle(ShownLifecycleScope())
   private var toolbarView: ToolbarView? = null
 
   fun init(toolbarView: ToolbarView, navigator: Navigator) {
     this.toolbarView = toolbarView
-    toolbarView.binding.back.setOnClickListener {
-      navigator.goBack()
-    }
+    toolbarView.binding.back.setOnClickListener { navigator.goBack() }
+    NavigationPropagator.addNavigableListener(this)
```
ryanmoelter, line 24: Should we add this listener in `onShow` for symmetry?

ryanmoelter (approved):
Overall looks good, thanks for fixing this!

magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/toolbar/ToolbarHelper.kt
cmathew: I think I had it this way to support the Activity being killed by the OS while backgrounded (as opposed to finished by the back button). But it's just a sample, and this makes the code much simpler

PR #211, opened 2021-10-07
containerView can be null when background app while navigating

ydxb7:


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -148,7 +148,7 @@ public class NavigationDelegate(
         NoAnimationTransition()
       }
       transition.animate(from, to, direction) {
-        if (context != null) {
+        if (context != null && containerView != null) {
```
cmathew, line 151: odd that context could be non-null, but containerView is null. Going to investigate this a bit

PR #221, opened 2021-11-01
Add LegacyStep

cmathew:
convenience wrapper to aid in migrating Screens to the Steps

Also adds a few tests to our "Migration" sample, for demonstration purposes


magellan-legacy/src/main/java/com/wealthfront/magellan/LegacyStep.kt
```diff
@@ -0,0 +1,23 @@
...
+import android.content.Context
+import android.view.View
+import androidx.annotation.VisibleForTesting
+import com.wealthfront.magellan.core.Navigable
+import com.wealthfront.magellan.lifecycle.LifecycleAwareComponent
+import com.wealthfront.magellan.lifecycle.attachFieldToLifecycle
+import com.wealthfront.magellan.lifecycle.createAndAttachFieldToLifecycleWhenShown
+import com.wealthfront.magellan.view.DialogComponent
+
+public abstract class LegacyStep<V : View>(protected val navigator: Navigator) :
```
cmathew, line 12: I am hoping that asking for a navigator argument will force developers to think about / maintain the backstack "scope" at which this LegacyStep operates

magellan-sample-migration/src/main/java/com/wealthfront/magellan/sample/migration/tide/HelpScreen.kt
```diff
@@ -3,27 +3,32 @@ package com.wealthfront.magellan.sample.migration.tide
...
-import com.wealthfront.magellan.sample.migration.SampleApplication.Companion.app
+import com.wealthfront.magellan.sample.migration.AppComponentContainer
 import com.wealthfront.magellan.sample.migration.api.DogApi
 import rx.android.schedulers.AndroidSchedulers
 import rx.schedulers.Schedulers
 import javax.inject.Inject
 
-class HelpScreen : Screen<HelpView>() {
+@OpenForMocking
+class HelpScreen(navigator: Navigator) : LegacyStep<HelpView>(navigator) {
```
cmathew, line 18: oops, I think I want LinearNavigator here..

magellan-sample-migration/src/test/java/com/wealthfront/magellan/sample/migration/tide/HelpScreenTest.kt
```diff
@@ -0,0 +1,70 @@
...
+          status = "something"
+        )
+      )
+    )
+  }
+
+  @Test
+  fun fetchesDogPicOnShow() {
+    helpScreen.transitionToState(LifecycleState.Created(context))
+    helpScreen.view = view
```
ryanmoelter, line 65: I don't think this will actually override the view, since it will get set again in `show()` when the `LegacyStep` calls `createView()`. There's no good way to fake views if we use `transitionToState` unfortunately

ryanmoelter (approved):
This should make migration *much* easier, thanks for making this!

magellan-sample-migration/src/test/java/com/wealthfront/magellan/sample/migration/tide/HelpScreenTest.kt
cmathew: great catch. For some reason this was passing locally for me (???)

Anyway I did a sorta hacky thing. Overriding HelpScreen#createView in test, kinda like a Spy

PR #228, opened 2021-12-22
[AP-3542] Support passing in custom shownScope

cmathew:
this could be useful in integration test settings, where we're interested in using a Test Dispatcher


cmathew:
TODO
- Update `private final ShownLifecycleScope shownScope = new ShownLifecycleScope();` in Screen.java
- Verify locally that the change affects integration tests in the desired way

PR #230, opened 2022-01-21
Add callback to override navigation intents

cmathew:
By returning true, `NavigationRequestHandler#onNavigationRequested` can block a navigation from happening. The idea is that `onNavigationRequested` can perform its own replacement navigation, show a dialog, whatever. 

Key use-case is applying some global logic around when users should be asked to update their app.


magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/SampleApplication.kt
```diff
@@ -12,6 +19,19 @@ class SampleApplication : Application() {
     appComponent = DaggerAppComponent.builder()
       .appModule(AppModule())
       .build()
+
+    Magellan.navigationRequestHandler = object: NavigationRequestHandler {
+      override fun onNavigationRequested(
+        navigationDelegate: NavigationDelegate,
+        navigable: NavigableCompat
+      ): Boolean {
+        if (navigable is SuggestExhibitJourney) {
```
cmathew, line 28: arbitrary example: this version of the app shouldn't be allowed to see SuggestExhibitJourney

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -63,6 +69,16 @@ public open class NavigationDelegate(
     direction: Direction,
     backStackOperation: (Deque<NavigationEvent>) -> MagellanTransition
   ) {
+    navigationRequestHandler?.let { navRequestHandler ->
+      val backstackCopy = ArrayDeque(backStack)
+      backStackOperation.invoke(backstackCopy)
```
cmathew, line 74: this will do a "dry run" of the navigation operation on a copy of the backstack, from which it obtains a candidate currentNavigable

ryanmoelter:
Hmm, I have to take more of a look at this before having strong opinions, but this seems like a kind of fragile way to do this.

Could we instead make a `VersionCheckerStep` that wraps another `Step`/`Journey`, takes an enum like a `VersionSensitiveFeature` or something, and does the check there? That way, the check is happening at the local level rather than needing to be maintained at a global level, meaning it's easier to remember to update it when e.g. you change what journey/step is associated with a feature, and it's one less thing that we configure through `Magellan`.

Another option would be to wrap that sort of version-checking logic in a `LifecycleAware` component of some sort, basically the same effect through a slightly different mechanism.

The downside would be that we have to do this at each entry point to a feature, but for new features at least that should be a `Journey` (i.e. only one entry point), so that shouldn't be too bad once we migrate more.

Obv nothing I say is a blocker, but I'll look at this again over the weekend or early next week. (Going skiing tomorrow! 🏂️)

kyeungk (approved):

cmathew:
> Hmm, I have to take more of a look at this before having strong opinions, but this seems like a kind of fragile way to do this.
> 
> Could we instead make a `VersionCheckerStep` that wraps another `Step`/`Journey`, takes an enum like a `VersionSensitiveFeature` or something, and does the check there? That way, the check is happening at the local level rather than needing to be maintained at a global level, meaning it's easier to remember to update it when e.g. you change what journey/step is associated with a feature, and it's one less thing that we configure through `Magellan`.
> 
> Another option would be to wrap that sort of version-checking logic in a `LifecycleAware` component of some sort, basically the same effect through a slightly different mechanism.
> 
> The downside would be that we have to do this at each entry point to a feature, but for new features at least that should be a `Journey` (i.e. only one entry point), so that shouldn't be too bad once we migrate more.
> 
> Obv nothing I say is a blocker, but I'll look at this again over the weekend or early next week. (Going skiing tomorrow! 🏂️)

PR #234, opened 2022-02-03
Support custom view creation operations at Navigator level

cmathew:
allows Navigators to manipulate views right after they are created. Application I have in mind is per-Navigator templating - perhaps we want to "wrap" all ScreenViews into a LinearLayout, with a per-Screen Toolbar sibling, before adding to the containerview.


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -149,10 +150,8 @@ public open class NavigationDelegate(
     navigationPropagator.onNavigatedTo(currentNavigable)
     when (currentState) {
       is LifecycleState.Shown, is LifecycleState.Resumed -> {
-        containerView!!.addView(
-          currentNavigable.view!!,
-          direction.indexToAddView(containerView!!)
-        )
+        val templatedView = templateApplier?.onViewCreated(currentNavigable.view!!) ?: currentNavigable.view!!
```
Kritarie, line 153: `navigateTo` and `navigateFrom` should return the templated view rather than the navigable's view

magellan-library/src/main/java/com/wealthfront/magellan/lifecycle/CreateAndAttachFieldToLifecycleWhenShownDelegate.kt
```diff
@@ -20,3 +21,40 @@ public class CreateAndAttachFieldToLifecycleWhenShownDelegate<Field>(
 
 public fun <Field> LifecycleOwner.createAndAttachFieldToLifecycleWhenShown(fieldSupplier: (Context) -> Field): AttachFieldToLifecycleDelegate<CreateAndAttachFieldToLifecycleWhenShownDelegate<Field>, Field?> =
   attachFieldToLifecycle(CreateAndAttachFieldToLifecycleWhenShownDelegate(fieldSupplier), { it.field })
+
+
+public class OverrideableShownFieldDelegate<PropertyType>(
```
cmathew, line 26: I couldn't use CreateAndAttachFieldToLifecycleWhenShownDelegate because it's built on top of AttachFieldToLifecycleDelegate, which ceases lifecycle listening whenever fields are overridden. Since I am planning to override templatedView (inside NavigatorDelegate), yet *also* want it to stay attached (so that it get torn down onHide, and initialized to a good default inside onShow) I wrote this relatively naive Delegate instead

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationDelegate.kt
```diff
@@ -149,20 +150,19 @@ public open class NavigationDelegate(
     navigationPropagator.onNavigatedTo(currentNavigable)
     when (currentState) {
       is LifecycleState.Shown, is LifecycleState.Resumed -> {
-        containerView!!.addView(
-          currentNavigable.view!!,
-          direction.indexToAddView(containerView!!)
-        )
+        val templatedView = templateApplier?.onViewCreated(currentState.context!!, currentNavigable.view!!) ?: currentNavigable.view!!
```
cmathew, line 153: this allows each navigator to customize UI for each Navigable it puts into the containerView. This is separate from navigator-level UI - to do that, one should continue to define UI in their Journey's XML.

magellan-library/src/main/java/com/wealthfront/magellan/core/Displayable.kt
```diff
@@ -6,6 +6,8 @@ public interface Displayable {
 
   public val view: View?
 
+  public var templatedView: View?
```
cmathew, line 9: storing any generated templatedView on the Displayable, to preserve encapsulation. I suppose it could also work to store a map at the navigator level (i.e. Displayable -> It's templated View, if there is one). But I had a couple of worries about that: 
- Could be harder to ensure the map gets cleaned up as needed when Displayable are torn down. I guess we'd do this inside navigateTo/navigateFrom instead of in response to onHide events? Feels more brittle.
- Need to guarantee unique keys for this map. However, it's currently enforced that a given Navigable only appears once in a given backstack so this likely isn't a problem.

cmathew: Possibly dumb question, is there a way to include an `internal` property in an interface? I'd really rather not expose `templatedView` property to developers working inside a Screen/Step etc.

magellan-sample-advanced/src/androidTest/java/com/wealthfront/magellan/IntegrationTestRunner.kt
```diff
@@ -0,0 +1,15 @@
+package com.wealthfront.magellan
+
+import androidx.test.runner.AndroidJUnitRunner
+import com.squareup.rx2.idler.Rx2Idler
+import io.reactivex.plugins.RxJavaPlugins
+
+class IntegrationTestRunner : AndroidJUnitRunner() {
+
+  override fun onStart() {
+    RxJavaPlugins.setInitIoSchedulerHandler(Rx2Idler.create("RxJava 2.x IO Scheduler"))
```
Kritarie, line 10: Fancy

Kritarie (approved):

PR #238, opened 2022-02-25
Conditionally restore backstack to undo unwanted operations

cmathew:
meanwhile if it's an operation we want, then just continue as normal.

The problem of the [existing "dry run" approach](https://github.com/wealthfront/magellan/pull/230/files#r789325119) is that it potentially runs all backstack operations twice - once for the "dry run", and again "for real". This causes problems if you have non-idempotent state changes inside a backstack operation (example: manipulating the intent on an Activity during deeplinking logic). 

With this approach, we're guaranteed to only run each backstack operation once.

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/init/Magellan.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/init/Magellan.kt
@@ -15,7 +15,7 @@ public object Magellan {
   internal var logDebugInfo: Boolean = false
   internal var disableAnimations: Boolean = false
   internal var customDefaultTransition: MagellanTransition = DefaultTransition()
-  public var navigationOverrides: Set<NavigationOverride> = emptySet()
+  internal var navigationOverrides: List<NavigationOverride> = emptyList()
 
   @JvmStatic
   @JvmOverloads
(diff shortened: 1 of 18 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/Navigator.kt
```diff
@@ -4,10 +4,16 @@ import android.content.Context
 import android.view.View
 
 public interface NavigationRequestHandler {
-  public fun onNavigationRequested(
+  public fun shouldOverrideNavigation(
     navigationDelegate: NavigationDelegate,
     navigable: NavigableCompat
   ): Boolean
+
+  // what if this supplied a backstack operation, instead?
```
cmathew, line 12: TODO as noted I might change the API here. Currently there's an implicit contract that `overrideNavigationRequest` will call navigationDelegate.navigate somewhere in its code..we can make that explicit by having `overrideNavigationRequest` supply an alternative backstack operation, instead.

magellan-library/src/main/java/com/wealthfront/magellan/init/Magellan.kt
```diff
@@ -14,23 +15,35 @@ public object Magellan {
   internal var logDebugInfo: Boolean = false
   internal var disableAnimations: Boolean = false
   internal var customDefaultTransition: MagellanTransition = DefaultTransition()
-  public var navigationRequestHandler: NavigationRequestHandler? = null
+  public var navigationOverrides: Set<NavigationOverride> = emptySet()
```
cmathew, line 18: hmm this should be a `list` instead of a `set`. If there are multiple overrides desired, we want to give developers an opportunity to order them

PR #241, opened 2022-05-06
Introduce LegacyJourney

cmathew:
...for better 1x and 2x interop.


Kritarie (approved):
LGTM as-is. Curious if it makes sense to have a `SimpleLegacyJourney` similar to `SimpleJourney`?

PR #242, opened 2022-06-14
Support whenTransitionFinished deferred callback in Steps

Kritarie:


cmathew (approved):

magellan-library/src/main/java/com/wealthfront/magellan/core/Step.kt
```diff
@@ -65,4 +69,14 @@ public abstract class Step<V : ViewBinding>(
   protected open fun onPause(context: Context, binding: V) {}
 
   protected open fun onHide(context: Context, binding: V) {}
+
+  @CallSuper
```
ryanmoelter, line 73: Historically, we've avoided needing `@CallSuper` by making the `override` `final` exposing an "onXyz" method. Is that a pattern we considered here?

PR #248, opened 2022-06-27
Define backPress behavior for LazySetNavigator

cmathew:
effectively this means that only the active child of LazySetNavigator will receive the back event


magellan-sample-advanced/src/main/java/com/wealthfront/magellan/sample/advanced/MainMenuStep.kt
```diff
@@ -74,6 +74,18 @@ class MainMenuStep(
     binding.bottomBarNavigation.selectedItemId = selectedTab
   }
 
+  override fun onBackPressed(): Boolean {
+    if (navigator.backPressed()) {
```
cmathew, line 78: TODO don't need this. I believe LifecycleComponent's backPressed already does what we want

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
```diff
@@ -129,5 +129,7 @@ public open class LazySetNavigator(
     }
   }
 
-  override fun onBackPressed(): Boolean = currentNavigable?.backPressed() ?: false
+  override fun backPressed(): Boolean {
```
cmathew, line 132: either we override LazySetNavigator.backPressed (instead of the usual onBackPressed), or we change LifecycleRegistry.onAllActiveListenersUntilTrue to filter out NO_LIMIT instead of SHOWN. I chose this option for being lower-impact

Issue #254, opened 2022-08-01
Scope currentScreen() method to Navigator it is called on


cmathew:
this method needn't be static if we could figure out a way to automatically propagate and attach navigation events from child journeys to parent ones

e.g. you could then inject the root expedition and call root.currentNavigable on it.

PR #255, opened 2022-08-04
currentNavigable

Kritarie:
Removes global CurrentNavigableProvider for recursive, scoped currentNavigable.

#254


magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableCompat.kt
```diff
@@ -3,4 +3,6 @@ package com.wealthfront.magellan.navigation
 import com.wealthfront.magellan.core.Displayable
 import com.wealthfront.magellan.lifecycle.LifecycleAware
 
-public interface NavigableCompat : LifecycleAware, Displayable
+public interface NavigableCompat : LifecycleAware, Displayable {
+  public val currentNavigable: NavigableCompat get() = this
```
cmathew, line 7: hmm why is this method on Navigable instead of Navigator? Please don't tell me it's because of LazySetNavigator 😩

cmathew: oh I see - it lets us avoid reflection when sifting thru backstacks. We can just treat all Navigables equally i.e. without checking for Navigators only

cmathew (approved):

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigableCompat.kt
cmathew: oh, and this also allows the application developer to define current screen when it's otherwise undefined. Example: a Step containing two "sibling" Journeys doesn't have a well-defined "current" screen

Kritarie: Exactly. I could imagine a master-detail arrangement, where an application developer would have a better sense of what the "current" navigable is

PR #256, opened 2022-08-08
Update Kotlin + Gradle plugins

cmathew:
Kotlin to 1.6.20
AGP to 7.2.2

- minor style changes
- update optIn annotations and flags


magellan-library/src/test/java/com/wealthfront/magellan/coroutines/CreatedLifecycleScopeTest.kt
```diff
@@ -41,86 +38,77 @@ internal class CreatedLifecycleScopeTest {
   }
 
   @Test
-  fun cancelBeforeCreated() {
-    runBlockingTest {
+  fun cancelBeforeCreated() = runTest {
+    val async = createdScope.async { delay(5000) }
+    assertThat(async.isCancelled).isTrue()
+    assertThat(async.getCancellationException()).hasMessageThat().contains("Not created yet")
```
cmathew, line 44: there is probly a better way to inspect these versus using `InternalCoroutinesApi` (try-catch block?). But leaving it to keep this PR simple as possible

PR #257, opened 2022-08-09
Trigger afterNavigation event only after any animated transitions are complete

cmathew:
NavigationPropagator events summarized:
- `beforeNavigation` is fired at the very beginning of the navigation process
- `afterNavigation` currently fires after both to/from navigable lifecycle transitions have occurred. This PR changes that to wait until the transition animations are complete and the view is updated
- `onNavigatedTo` is called after the "to" navigable gets ramped up to RESUMED
- `onNavigatedFrom` is called after the "from" navigable is ramped down to CREATED


cmathew:
TODO unit testing

PR #259, opened 2022-08-10
Customizing NavigationTraverser

Kritarie:
Allow application developers to customize navigation traverser behavior by overriding createSnapshot. This is useful for more advanced navigation use-cases like master-detail or tabbed navigables.

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
@@ -7,28 +7,12 @@ import com.wealthfront.magellan.debug.INDENT_SPACE
 import com.wealthfront.magellan.debug.VERTICAL_LINE
 import com.wealthfront.magellan.debug.VERTICAL_T
 import com.wealthfront.magellan.init.shouldLogDebugInfo
-import com.wealthfront.magellan.lifecycle.LifecycleOwner
-import java.util.ArrayDeque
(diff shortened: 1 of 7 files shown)


magellan-library/src/test/java/com/wealthfront/magellan/navigation/NavigationTraverserTest.kt
```diff
@@ -180,4 +210,15 @@ internal class NavigationTraverserTest {
 
   private inner class DummyStep4 :
     Step<MagellanDummyLayoutBinding>(MagellanDummyLayoutBinding::inflate)
+
+  private inner class CustomStep(val customChild: NavigableCompat) : Step<MagellanDummyLayoutBinding>(MagellanDummyLayoutBinding::inflate) {
```
cmathew, line 214: thanks for testing arbitrary overrides

cmathew (approved):

magellan-library/src/main/java/com/wealthfront/magellan/navigation/NavigationTraverser.kt
```diff
@@ -71,22 +55,22 @@ public class NavigationTraverser(private val root: NavigableCompat) {
   }
 }
 
-public data class NavigationNode(val value: NavigableCompat, var children: List<NavigationNode>) {
-
-  val hasBackstack: Boolean
```
cmathew, line 76: nice to get rid of this "reflection" in favor of explicit overrides at the Navigable level

PR #260, opened 2022-08-11
Support transition interrupt

cmathew:
- adds a required `interrupt` method to each transition class. In all existing classes, implement this by `end`-ing the animation
- trigger interrupts as needed inside `LazySetNavigator`
- bonus: Add circular reveal transition somewhere in the sample apps


magellan-library/src/main/java/com/wealthfront/magellan/transitions/CrossfadeTransition.kt
```diff
@@ -2,31 +2,58 @@ package com.wealthfront.magellan.transitions
...
+import android.animation.AnimatorSet
+import android.animation.ObjectAnimator
+import android.util.Property
 import android.view.View
 import android.view.View.GONE
 import com.wealthfront.magellan.Direction
 
 public class CrossfadeTransition : MagellanTransition {
 
+  private var animationSet: Animator? = null
```
cmathew, line 14: the existing CrossfadeTransition worked by fading out, and then fading in sequentially. I'm not sure why that was, but this will do both simultaneously as it's simpler to cancel that

Kritarie (approved):
Perhaps beyond the scope of this PR, but does `DefaultLinearNavigator` need to support transition interrupts as well?

magellan-library/src/main/java/com/wealthfront/magellan/transitions/CrossfadeTransition.kt
cmathew: err wait a sec. AnimatorSet has a `playSequentially` command. Going to use that

magellan-library/src/main/java/com/wealthfront/magellan/transitions/CircularRevealTransition.kt
```diff
@@ -19,6 +19,12 @@ import kotlin.math.hypot
  */
 public class CircularRevealTransition(private val clickedView: View) : MagellanTransition {
 
+  private var animator: Animator? = null
+
+  override fun interrupt() {
+    animator?.end()
```
cmathew, line 25: sadly wasn't able to figure out a unit test for this. Couldn't get Robolectric to consider the necessary views attached

PR #264, opened 2022-09-21
Adds several navigation helper shorthand functions to LinearNavigator.

Kritarie:
API provides:
`goBackTo(navigable)`
`goBackBefore(navigable)`
`goBackToRoot()`
`goForwardToRoot()`
`resetWithRoot(navigable)`
`clearUntilRootAndGoTo(navigable)`

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigatorExtensions.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigatorExtensions.kt
@@ -1,0 +1,84 @@
+package com.wealthfront.magellan.navigation
+
+import com.wealthfront.magellan.Direction
+import com.wealthfront.magellan.core.Navigable
+import com.wealthfront.magellan.init.getDefaultTransition
+import com.wealthfront.magellan.transitions.MagellanTransition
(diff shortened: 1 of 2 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigatorExtensions.kt
```diff
@@ -0,0 +1,84 @@
...
+    transition
+  }
+}
+
+public fun LinearNavigator.clearUntilRootAndGoTo(
+  navigable: Navigable,
+  overrideTransition: MagellanTransition? = null
+) {
+  navigate(Direction.FORWARD) { backStack ->
+    while (backStack.size > 1) {
```
cmathew, line 59: minor: worth adding a  `backStack.isEmpty()` check?

cmathew (approved):

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LinearNavigatorExtensions.kt
Kritarie: done!

PR #275, opened 2023-04-07
Fixes CircularRevealTransition

Kritarie:


magellan-library/src/main/java/com/wealthfront/magellan/transitions/CircularRevealTransition.kt
```diff
@@ -7,17 +7,23 @@ import android.os.Build
...
 /**
  * A [MagellanTransition] that reveals the next screen circularly outward from the middle of the
- * [clickedView].
+ * [clickedViewId].
  *
- * @property clickedView the view on which this circular reveal is centered
+ * @property clickedViewId the id of the view on which this circular reveal is centered
  */
-public class CircularRevealTransition(private val clickedView: View) : MagellanTransition {
+private class CircularRevealTransition(@IdRes private val clickedViewId: Int) : MagellanTransition {
```
Kritarie, line 26: Hold the id instead of the view to avoid leak

magellan-library/src/main/java/com/wealthfront/magellan/transitions/CircularRevealTransition.kt
```diff
@@ -31,33 +37,38 @@ public class CircularRevealTransition(private val clickedView: View) : MagellanT
...
-        interpolator = FastOutSlowInInterpolator()
-      }
-      animator!!.start()
-    } else {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
       onAnimationEndCallback()
+      return
     }
+
+    val clickedViewContainer = if (direction == Direction.FORWARD) from else to
```
Kritarie, line 45: Support back animation properly. We're assuming that the clicked view is inside the "from" container if we're going forward, or inside the "to" container if we're going backward.

magellan-library/src/main/java/com/wealthfront/magellan/transitions/CircularRevealTransition.kt
cmathew: nice. Sure, I guess we can assume the ID will be unique

cmathew (approved):

PR #283, opened 2023-09-22
Use Android Gradle plugin v7.4.2

cmathew:


internal-test-support/build.gradle
```diff
@@ -7,8 +7,6 @@ android {
   defaultConfig {
     minSdkVersion Versions.minSdkVersion
     targetSdkVersion Versions.targetSdkVersion
-    versionCode 1
-    versionName "1.0"
```
cmathew, line 11: these versionCode and versionName are unused

PR #287, opened 2023-09-27
Basic unit tests for all sample-migration screens

cmathew:
- mock toolbar helper in tests
- use setContentScreen where possible
- configure Robolectric via properties file where possible

### DIFF
--- a/magellan-sample-migration/src/main/java/com/wealthfront/magellan/sample/migration/toolbar/ToolbarHelper.kt
+++ b/magellan-sample-migration/src/main/java/com/wealthfront/magellan/sample/migration/toolbar/ToolbarHelper.kt
@@ -2,6 +2,7 @@ package com.wealthfront.magellan.sample.migration.toolbar
 
 import android.content.Context
 import com.wealthfront.magellan.Navigator
+import com.wealthfront.magellan.OpenForMocking
 import com.wealthfront.magellan.lifecycle.LifecycleAwareComponent
 import com.wealthfront.magellan.navigation.NavigationListener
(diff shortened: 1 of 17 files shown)


magellan-sample-migration/src/test/resources/robolectric.properties
```diff
@@ -1 +1,2 @@
-sdk=28
\ No newline at end of file
+sdk=28
+application=com.wealthfront.magellan.sample.migration.TestSampleApplication
```
cmathew, line 2: Robolectric will pick this up automatically since it's looking out for classes named "TestBlahApplication", but good to be explicit anyways

magellan-sample-migration/src/main/java/com/wealthfront/magellan/sample/migration/toolbar/ToolbarHelper.kt
```diff
@@ -12,7 +13,8 @@ import com.wealthfront.magellan.navigation.NavigationPropagator
  * Ideally, this dependency is provider by dependency injection and configured so that the views are cleaned up when the activity
  * is destroyed with the help of (subcomponents & custom scopes)[https://dagger.dev/dev-guide/subcomponents].
  */
-object ToolbarHelper : LifecycleAwareComponent(), NavigationListener {
+@OpenForMocking
+class ToolbarHelper : LifecycleAwareComponent(), NavigationListener {
```
cmathew, line 17: convert to instance so it can be mocked in test

kyeungk (approved):

PR #290, opened 2023-10-27
Use constructor injection in sample app

cmathew:
this will make providing mocks more straightforward in tests


magellan-sample-migration/src/test/java/com/wealthfront/magellan/sample/migration/tide/DogListStepTest.kt
```diff
@@ -17,22 +16,20 @@ import org.junit.runner.RunWith
...
   private val activityController = Robolectric.buildActivity(ComponentActivity::class.java)
 
   @Before
   fun setUp() {
     val context = ApplicationProvider.getApplicationContext<Application>()
-    ((context as AppComponentContainer).injector() as TestAppComponent).inject(this)
-
-    coWhen { api.getAllBreeds() }
+    val component = ((context as AppComponentContainer).injector() as TestAppComponent)
+    dogListStep = DogListStep(component.toolbarHelper, component.api) { chosenBreed = it }
```
Kritarie, line 31: Instead of exposing individual services on TestAppComponent, what do you think about having Dagger inject DogListStepTest with a DogListStepFactory?

Kritarie (approved):

PR #293, opened 2023-10-30
Support backstack read operations from a background thread

Kritarie:


magellan-library/src/main/java/com/wealthfront/magellan/navigation/CopyOnWriteArrayDeque.kt
cmathew, line 1: your goal is to use a thread-safe Deque? Does this class work for you: https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ConcurrentLinkedDeque.html

Kritarie: I think it would work, but we would need to bump the min sdk version to API 21. That's probably fine, right?
Screenshot 2023-10-31 at 12 21 36 PM

PR #299, opened 2024-04-24
Don't use version catalog for project level plugin mgmt

cmathew:
to avoid confusing limitations and issues

rolls back some changes https://github.com/wealthfront/magellan/pull/297/files. Also, adds kotlin-gradle-plugin to buildSrc to ensure we're using the desired Kotlin version. This is especially important for the Compose toolchain

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-  id 'org.jetbrains.kotlin.jvm' version '1.8.21'
+  id 'org.jetbrains.kotlin.jvm' version '1.8.22'
 }
 
 repositories {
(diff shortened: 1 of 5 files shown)


buildSrc/build.gradle
```diff
@@ -10,6 +10,7 @@ repositories {
 
 dependencies {
   implementation 'com.android.tools.build:gradle:8.1.2'
+  implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.8.22'
```
cmathew, line 13: without this, we get many strange errors in Compose compilation. Thanks to https://stackoverflow.com/questions/75995893/jetpack-compose-most-basic-buildsrc-build-error-illegalstateexception-coul

PR #304, opened 2024-10-23
Support removing navigables from LazySetNavigator

gerin98:
Adding support for `LazySetNavigator` to remove navigables
- Can remove a single Navigable
- Can remove multiple Navigables
- Can update the navigator's Navigables which can add and remove Navigables

### DIFF
--- a/magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
+++ b/magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
@@ -23,6 +23,9 @@ public open class LazySetNavigator(
   private var ongoingTransition: MagellanTransition? = null
 
   @VisibleForTesting
+  internal var existingNavigables: MutableSet<NavigableCompat> = mutableSetOf()
+
+  @VisibleForTesting
   internal var containerView: ScreenContainer? = null
   private var currentNavigable: NavigableCompat? = null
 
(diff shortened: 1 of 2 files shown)


magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
```diff
@@ -36,9 +39,39 @@ public open class LazySetNavigator(
...
+    }
+  }
+
+  public fun updateNavigables(navigables: Set<NavigableCompat>, handleCurrentTabRemoval: () -> Unit) {
+    val navigablesToRemove = existingNavigables subtract navigables
+    val navigablesToAdd = navigables subtract existingNavigables
+
+    if (navigablesToRemove.contains(currentNavigable)) {
+      handleCurrentTabRemoval()
+    }
```
gerin98, line 69: There's an edge case to be aware of when removing a Navigable that is currently visible. We need to switch to another navigable before we can remove it because once you call `removeFromLifecycle` on a navigable you can't call `replace` on it anymore. I wrote a test `updateMultipleNavigables_andRemoveCurrentNavigable` describing this case.

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
```diff
@@ -36,9 +39,39 @@ public open class LazySetNavigator(
   }
 
   public fun addNavigable(navigable: NavigableCompat) {
+    existingNavigables.add(navigable)
     lifecycleRegistry.attachToLifecycleWithMaxState(navigable, LifecycleLimit.CREATED)
   }
 
+  public fun removeNavigables(navigables: Set<NavigableCompat>) {
+    for (navigable in navigables) {
```
Kritarie, line 47: Is there a risk of `ConcurrentModificationException` if one of the navigables tries to add / remove a navigable in its `onDestroy` ? I don't know enough about `mutableSetOf` to know if it allows concurrent modification

magellan-library/src/main/java/com/wealthfront/magellan/navigation/LazySetNavigator.kt
```diff
@@ -36,9 +41,44 @@ public open class LazySetNavigator(
...
+      removeNavigable(navigable)
+    }
+  }
+
+  @RequiresApi(Build.VERSION_CODES.N)
+  public fun removeNavigable(navigable: NavigableCompat) {
+    existingNavigables.removeIf { it == navigable }
+    if (lifecycleRegistry.children.contains(navigable)) {
+      lifecycleRegistry.removeFromLifecycle(navigable)
```
gerin98, line 59: `removeIf` avoids `ConcurrentModificationException` but requires API level 24

Kritarie (approved):

kyeungk (approved):

